PR #20, opened 2019-10-03
Fix files provider to search tools in order they are listed

tkonolige:
The files provider used a dict to organize the tools it was searching. Iterating over the dict did not give items in the same order as they appeared in the file. This caused `git` to be used before `fd`.

This PR fixes the issue by switching from a dict to a list.


liuchengxu:
I'm actually seeing this problem too and fix it in #18, but thanks anyway! 🍻

Issue #34, opened 2019-10-09
Commands Ctrl+t and friends throw error


alerque:
The shortcuts for opening files in a new tab or split return an error.

 - OS: Linux (Arch)
 - NeoVim version: 0.4.2
 - vim-clap version: current HEAD
 - Have you reproduced with a minimal vimrc: yes

`:Clap! files` to launch the files provider, select any file, and hit <kbd>Ctrl</kbd><kbd>t</kbd> to open file in a new tab. The following error is thrown:

```txt
Error detected while processing function <SNR>39_action:
line 2:
E117: Unknown function: <SNR>39_sink
```

liuchengxu:
<img width="921" alt="屏幕快照 2019-10-09 下午4 03 29" src="https://user-images.githubusercontent.com/8850248/66462794-67e85800-eaae-11e9-997c-7e34a589ccfb.png">

It's still in an unimplemented state.

alerque:
@liuchengxu That is not very clear from the documentation. Other things marked that way _are_ implemented (e.g. <kbd>Ctrl</kbd><kbd>h</kbd>). It would be a lot more clean if ⓐ it was up to date and ⓑ the _completed_ items used the same Markdown notation, marking the boxes as checked (`- [x]`). Mixing and matching notations leaves the intention unclear.

liuchengxu:
Feel free to improve it :(.

alerque:
Will do. Lets also repurpose this issue as a feature request for those things since it's a bit easier to track here than in a README ;-)

numbcoder:
> <img alt="屏幕快照 2019-10-09 下午4 03 29" width="921" src="https://user-images.githubusercontent.com/8850248/66462794-67e85800-eaae-11e9-997c-7e34a589ccfb.png">
> 
> It's still in an unimplemented state.

When will this feature be implemented? It's so useful

liuchengxu:
No ETA. A lot of more important things need to be done.

alerque:
@liuchengxu I understand everybody has different priorities, but in evaluating this for my own use this was the only thing I found that was a blocker to migrate from my `fzf-tmux` setup which has this feature. Everything else I actively use was on par or better with what.

liuchengxu:
I have raised the priority of this issue in the favor of the requests. Please have a try with #39 and let me know if that doesn't work. @alerque @numbcoder

PR #36, opened 2019-10-09
Use todo-list syntax instead of mix and match lists

alerque:
This addresses the documentation ambiguity mentioned in #34.

I was unsure whether to mark off the <kbd>Ctrl</kbd><kbd>h</kbd> bit as done. It works for me in NeoVim but perhaps it isn't working everywhere? I added that as a separate commit so it would be easy to drop if it's not appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,17 +38,14 @@ Vim-clap is a modern generic interactive finder and dispatcher, based on the new
 
 ## Features
 
-- Pure vimscript.
-- Work out of the box, without any extra dependency.
-- Extensible, easy to add new source providers.
-- Find or dispatch anything on the fly, with smart cache strategy.
-- Untouch your current window layout, less eye movement.
-- Support multi-selection, use vim's regexp as filter by default.
-- Support the preview functionality when navigating the result list.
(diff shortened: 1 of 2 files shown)


liuchengxu:
```diff
--- a/autoload/clap/popup.vim
+++ b/autoload/clap/popup.vim
@@ -327,7 +327,7 @@ let s:move_manager["\<C-F>"] = s:move_manager.ctrl_f
 let s:move_manager["\<Right>"] = s:move_manager.ctrl_f
 let s:move_manager["\<C-E>"] = s:move_manager.ctrl_e
 let s:move_manager["\<BS>"] = s:move_manager.bs
-let s:move_manager["\<C-D>"] = s:move_manager.bs
+let s:move_manager["\<C-H>"] = s:move_manager.bs
 let s:move_manager["\<C-G>"] = s:move_manager.ctrl_g
```

@alerque  Can you please include this patch in your PR? Then I believe we could mark CTRL-H as done confidently. I just have not got enough time to take care of these details right now :).

alerque:
@liuchengxu Sure, done.

liuchengxu:
Thanks!

PR #70, opened 2019-10-19
Formalize context args

### LINKED ISSUE #28: Hidden files are not included in `:Clap files`
**Expected behavior**
Hidden files are included in the list of files returned

liuchengxu:
Close #28


ratheesh:
Regarding the implementation of `:Clap files --hidden`, instead of passing `--hidden` to individual commands, won't be useful to provide a global option like `g:Clap_include_hidden_files` flag so that it may be applied to all the commands?
The user is likely to decide about using hidden files globally rather than per command.

liuchengxu:
`--hidden` only applys to the grep provider. I'm not a fan of intoducing too many options when you already have a way out, thus I won't add such an option by myself.

PR #74, opened 2019-10-21
Implement sink for the default (underscore) provider

wookayin:
This simple implementation for invoking clap with selected clap provider works for me --- but any feedback or corner case would be appreciated!

### DIFF
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -74,7 +74,7 @@ endfunction
 
 function! s:_sink(selected) abort
   " a sink for "Clap _" (dispatch to other builtin clap providers).
-  call timer_start(0, {-> clap#(0, a:selected)})
+  call timer_start(0, {-> clap#_for(a:selected)})
 endfunction
 
 function! clap#_exit() abort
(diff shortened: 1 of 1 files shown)


liuchengxu:
Thanks for this patch! I was wondering how to make it work in a simple way and did not expect that using a timer is a solution. Works for me. After merging this, we could also add a description field to each provider so that we can know them well.

wookayin:
Thanks. Please make sure that we are using the correct bang option (that I wasn't 100% sure of).

liuchengxu:
This should correct the bang issue.

```diff
 function! s:_sink(selected) abort
   " a sink for "Clap _" (dispatch to other builtin clap providers).
-  call timer_start(0, {-> clap#(0, a:selected)})
+  call timer_start(0, {-> clap#(s:bang, a:selected)})
 endfunction
 
 function! clap#_exit() abort
@@ -254,6 +254,7 @@ function! clap#(bang, ...) abort
     call s:parse_opts(a:000[1:])
   endif
 
+  let s:bang = a:bang
   call clap#for(provider_id_or_alias)
 endfunction
```

liuchengxu:
🍻

wookayin:
Thanks!

PR #83, opened 2019-10-23
Add vint

liuchengxu:


alerque:
By coincidence I just setup Vint as a CI job on another vim plugin yesterday, but using GitHub Actions. If you happen to be interested in that instead of Travis feel free to rip off the implementation [from here](https://github.com/majutsushi/tagbar/blob/master/.github/workflows/vint.yml). The end to end run time for the CI job is about 20s to fire up the VM, install Python and Vint, lint the plugin, and return.

liuchengxu:
@alerque Thanks, I'm going to steal it then :(. And good to see you take the maintainance of tagbar :+1: .

.travis.yml
```diff
@@ -1,4 +1,21 @@
 sudo: required
```
alerque: Maybe &lt;del&gt; this file too? You probably don't need both. Note that the GA one won't run on _this PR_, the workflow has to land on master before it considers running it on PRs.

liuchengxu: Yeah, will drop that when everything is ready.

PR #86, opened 2019-10-23
Add extra (optional) arguments for grep

Olical:
This allows the user to enable things such as searching hidden files and folders with ripgrep.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ The option naming convention for provider is `g:clap_provider_{provider_id}_{opt
 - `g:clap_provider_grep_delay`: 300ms by default, delay for actually spawning the grep job in the background.
 
 - `g:clap_provider_grep_blink`: [2, 100] by default, blink 2 times with 100ms timeout when jumping the result. Set it to [0, 0] to disable the blink.
+- `g:clap_provider_grep_opts`: An empty string by default, allows you to enable flags such as `--hidden`.
 
 See `:help clap-options` for more information.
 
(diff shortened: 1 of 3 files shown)


Olical:
This is essentially lifted from the fzf + rg config I had, I used the exact same combination of `--hidden -g...`, it works super well for my use case. I thought adding a generic hook into the opts was a better options than a boolean flag that enabled hidden files.

This opens up further possibilities for people that have interesting requirements for their grepping. I hope it looks okay to you, totally understand if you don't like the change :slightly_smiling_face:

Let me know if you'd like anything tweaked! Thank you for the wonderful project, it's so damn good.

liuchengxu:
I'm ok with it. And @ratheesh should like this option :(. One concern is we currently only supports rg and that's ok. If we support the other tools later, e.g., ag, is this option is still clear enough?

ratheesh:
@Olical @liuchengxu Thanks a lot for this PR :-)

Olical:
A valid concern, @liuchengxu! I think if support for Ag or GNU Grep were
added it'd still be valid but the documentation would have to be updated to
reflect the contextual nature.

Since the option is user configured it should really be set alongside their
choice of grep program, they can choose the options they want for their
specific tool.

I was wondering if maybe the "rg -H..." full command should be configurable
but that requires a larger change. I'm of the opinion that rg is the best
tool for the job right now though, so I'm totally fine with building around
that for now :D

liuchengxu:
>I was wondering if maybe the "rg -H..." full command should be configurable
but that requires a larger change.

It could be and the files provider already supports the fully configurable choice using the context and provider args. But still, I'm not against if someone adds a global option for that.

Issue #76, opened 2019-10-21
Add a provider for yank history


ronisbr:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
It would be very nice to have a provider that shows the Yank history more or less to what this plugin does: https://github.com/vim-scripts/YankRing.vim

**Describe alternatives you've considered**
I am currently using YankRing, but it will be a very nice feature in vim-clap. Like, `Clap yank` (or something) will open a list of past yanked text so that you can select and paste by pressing enter.

**Additional context**
N/A

PR #87, opened 2019-10-24
[RFC]Add yanks provider

ratheesh:
Add yanks provider
This is largely based on https://github.com/sgur/ctrlp-extensions.vim

This should address #76 

I am new to vim scripting. Suggestions are welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ Command                                | List                               | Re
 `Clap lines`                           | Lines in the loaded buffers        | _none_
 `Clap marks`                           | Marks                              | _none_
 `Clap tags`                            | Tags in the current buffer         | **[vista.vim][vista.vim]**
+`Clap yanks`                           | Yank stack of the current vim session    | _none_
 `Clap windows` **<sup>!</sup>**        | Windows                            | _none_
 
 [fd]: https://github.com/sharkdp/fd
--- a/autoload/clap/provider/yanks.vim
+++ b/autoload/clap/provider/yanks.vim
@@ -1,0 +1,53 @@
+" Author: Ratheesh S <ratheeshreddy@gmail.com>
+" Description: List the recently yanked/deleted lines
+" Based on : https://github.com/sgur/ctrlp-extensions.vim
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+let s:yank_history = []
+let s:yanks = {}
+let s:max_yanks = get(g:, 'clap_provider_yanks_max_entries', 20)
+
+function! clap#provider#yanks#collect() abort
+  let last_yanked = getreg('"')
+
+  if !empty(s:yank_history) && last_yanked == s:yank_history[0]
+    return
+  endif
+
+  call filter(s:yank_history, 'v:val != last_yanked')
+  call insert(s:yank_history, last_yanked)
+
+  " Trim yank entries(purge old ones)
+  if len(s:yank_history) > s:max_yanks
+    call remove(s:yank_history, s:max_yanks, -1)
+  endif
+
+endfunction
+
+function! clap#provider#yanks#init() abort
+  augroup ClapYanksCollect
+    autocmd!
+    autocmd TextYankPost * call clap#provider#yanks#collect()
+  augroup END
+
+  " collect the data from default register
+  call clap#provider#yanks#collect()
+endfunction
+
+function! s:yanks.source() abort
+  return s:yank_history
+endfunction
+
+function! s:yanks_sink(selected) abort
+  call setreg('"', a:selected)
+  normal! ""p
+endfunction
+
+let s:yanks.sink = function('s:yanks_sink')
+
+let g:clap#provider#yanks# = s:yanks
+
+let &cpo = s:save_cpo
+unlet s:save_cpo
--- a/plugin/clap.vim
+++ b/plugin/clap.vim
@@ -19,3 +19,11 @@ augroup ClapBuffers
   autocmd BufWinEnter,WinEnter * let g:__clap_buffers[bufnr('')] = reltimefloat(reltime())
   autocmd BufDelete * silent! call remove(g:__clap_buffers, expand('<abuf>'))
 augroup END
+
+" yanks provider
+if get(g:, 'clap_enable_yanks_provider', 1)
+  augroup ClapYanks
+    autocmd!
+    autocmd VimEnter * call clap#provider#yanks#init()
+  augroup END
+endif
--- a/doc/clap.txt
+++ b/doc/clap.txt
@@ -23,7 +23,8 @@ CONTENTS                                                         *clap-contents*
     5.1 Highlights........................|clap-highlights|
   6. Provider Options.....................|clap-provider-options|
     6.1 Options for grep..................|clap-grep-options|
-    6.2 Other Provider Options............|clap-other-provider-options|
+    6.2 Options for yanks.................|clap-yanks-options|
+    6.3 Other Provider Options............|clap-other-provider-options|
   7. Commands.............................|clap-commands|
   8. Keybindings..........................|clap-keybindings|
   9. API..................................|clap-api|
@@ -390,9 +391,25 @@ g:clap_provider_grep_enable_icon              *g:clap_provider_grep_enable_icon*
 
   Set this variable to `0` to disable drawing an icon for the returned result.
 
+-------------------------------------------------------------------------------
+6.2. Clap Yanks Options                                      *clap-yanks-options*
+
+g:clap_enable_yanks_provider                       *g:clap_enable_yanks_provider*
+
+  Type: |Number|
+  Default: `1`(enabled)
+
+  Set this variable to `0` to disable yanks provider.
+
+g:clap_provider_yanks_max_entries             *g:clap_provider_yanks_max_entries*
+
+  Type: |Number|
+  Default: `20`
+
+  Set this variable to limit maximum yanks entries.
 
 -------------------------------------------------------------------------------
-6.2 Other Provider Options                          *clap-other-provider-options*
+6.3 Other Provider Options                          *clap-other-provider-options*
 
 
 ===============================================================================


README.md
```diff
@@ -94,6 +94,7 @@ Command                                | List                               | Re
 `Clap lines`                           | Lines in the loaded buffers        | _none_
 `Clap marks`                           | Marks                              | _none_
 `Clap tags`                            | Tags in the current buffer         | **[vista.vim][vista.vim]**
+`Clap yanks`                           | Yank stack of the current vim session    | ViM(8.0.1394+)/NeoVim(0.3.0+)
```
liuchengxu: I believe the requirement has been covered by clap itself.

ratheesh: Yes, adjusted accordingly.

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,70 @@
...
+endfunction
+
+function! s:yanks_sink(selected) abort
+    call setreg('"', a:selected)
+    normal! ""p
+endfunction
+
+let s:yanks.sink = function('s:yanks_sink')
+
+function! s:yanks.on_enter() abort
```
liuchengxu: The `on_enter` hook is unccessary as the collection is done via the TextYankPost autocmd already. It's often used in this plugin to initialize the filetype of display buffer to get the syntax highlight.

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,70 @@
...
+endfunction
+
+function! s:yanks_collect() abort
+    let yankstr = getreg('"')
+
+    if !empty(s:yanklist) && yankstr == s:yanklist[0]
+        return
+    endif
+
+    call s:remove_duplicated(s:yanklist, yankstr)
```
liuchengxu: Could be simpified using `filter`, `:h filter()`.

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,70 @@
...
+
+function! s:yanks_sink(selected) abort
+    call setreg('"', a:selected)
+    normal! ""p
+endfunction
+
+let s:yanks.sink = function('s:yanks_sink')
+
+function! s:yanks.on_enter() abort
+    if !exists('##TextYankPost')
```
liuchengxu: This check is also needless as the requirement of clap ensures the feature exists.

autoload/clap/provider/yanks.vim
ratheesh: `on_enter` hook is removed

autoload/clap/provider/yanks.vim
ratheesh: replaced function with `filter()` call as suggested.

autoload/clap/provider/yanks.vim
ratheesh: Removed the redundant part.

plugin/yanks.vim
```diff
@@ -0,0 +1,16 @@
+" vim-clap - Modern interactive filter and dispatcher
```
liuchengxu: Is it better to put this files in plug/clap.vim?

```vim
if get(g:, 'clap_enable_yanks_provider', 1)
  augroup ClapYanks
    autocmd!
    autocmd VimEnter * call clap#provider#yanks#init()
  augroup END
endif
```

ratheesh: Done!

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,46 @@
...
+" Based on : https://github.com/sgur/ctrlp-extensions.vim
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+let s:yanklist = []
+let s:yanks = {}
+
+function! s:yanks_collect() abort
+    let yankstr = getreg('"')
```
liuchengxu: Use 2 spaces for every level of indentation in Vim code.

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,46 @@
...
+    if !empty(s:yanklist) && yankstr == s:yanklist[0]
+        return
+    endif
+
+    call filter(s:yanklist, 'v:val != yankstr')
+    call insert(s:yanklist, yankstr)
+endfunction
+
+function! clap#provider#yanks#yanks_init() abort
+    augroup clap_yanklist
```
liuchengxu: The augroup name should be CamelCase, e.g, ClapYanksCollect.

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,46 @@
...
+
+    if !empty(s:yanklist) && yankstr == s:yanklist[0]
+        return
+    endif
+
+    call filter(s:yanklist, 'v:val != yankstr')
+    call insert(s:yanklist, yankstr)
+endfunction
+
+function! clap#provider#yanks#yanks_init() abort
```
liuchengxu: Personally I would rename to `clap#provider#yanks#init()` :(.

liuchengxu:
Overall good, just some nits. And I also have some ideas for the future improments:

- record the time of inserted history.
- add limit to the yanks history, e.g., keep up to 100 entries.
- save the yanks history to a local file so that we are able to use them when reopening vim.

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,46 @@
+" Author: Ratheesh S <ratheeshreddy@gmail.com>
+" Description: List the recently yanked/deleted lines
+" Based on : https://github.com/sgur/ctrlp-extensions.vim
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+let s:yanklist = []
```
liuchengxu: `s:yanks_history` is a better name?

autoload/clap/provider/yanks.vim
liuchengxu: `last_yanked` a better name?

plugin/clap.vim
```diff
@@ -19,3 +19,11 @@ augroup ClapBuffers
   autocmd BufWinEnter,WinEnter * let g:__clap_buffers[bufnr('')] = reltimefloat(reltime())
   autocmd BufDelete * silent! call remove(g:__clap_buffers, expand('<abuf>'))
 augroup END
+
+" yanks provider
+if get(g:, 'clap_enable_yanks_provider', 1)
+  augroup ClapYanks
+    autocmd!
+    autocmd VimEnter * call clap#provider#yanks#init()
+  augroup END
```
liuchengxu: We could avoid sourcing the yanks provider on VimEnter by introducing a global variable, although the sourcing is actually cheap :(.

```vim
if get(g:, 'clap_enable_yanks_provider', 1)
  let g:__clap_last_yanked = getreg('"')
  augroup ClapYanks
    autocmd!
    autocmd TextYankPost * call clap#provider#yanks#collect()
  augroup END
endif
```

Initialize `s:yank_history` with `g:__clap_last_yanked`:


```vim
let s:yank_history = exists('g:__clap_last_yanked') ? [g:__clap_last_yanked] : []
```

Expose `s:yanks_collect()` to `clap#provider#yanks#collect()`, and then we can remove `clap#provider#yanks#init()`.

ratheesh: `getreg()` in the above code executes before reading of `viminfo/shada` file before the VimEnter event.  As a result, yank provider starts with an empty list(even if register `"` is nonempty). Hence we may need to live with VimEnter autocmd for now. 

I have exposed `s:yanks_collect()` to `clap#provider#yanks#collect()`
I have implemented a max. yank entries through global variable(Default:20)

ratheesh:
> record the time of inserted history.

What is the use case scenario of the yank timestamp?

> add a limit to the yanks history, e.g., keep up to 100 entries.

This is included in current PR(latest commit)

> save the yanks history to a local file so that we are able to use them when reopening vim.

We can do this in a separate PR.

alerque:
Being a clipboard manager is a complicated problem, and not one you will find a lot of agreement on. I **do not want** my yanks being written to a file. I'm pretty against them hitting my disk even in temp space actually. Please don't hold up this PR for stuff like that. Having easy visual access to yanks will be awesome. Trying to much about with how yanks work in the first place should be out of scope for this.

autoload/clap/provider/yanks.vim
```diff
@@ -0,0 +1,53 @@
+" Author: Ratheesh S <ratheeshreddy@gmail.com>
+" Description: List the recently yanked/deleted lines
+" Based on : https://github.com/sgur/ctrlp-extensions.vim
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+let s:yank_history = []
+let s:yanks = {}
+let s:max_yanks = get(g:, 'clap_yanks_max_entries', 20)
```
liuchengxu: Please use `clap_provider_yanks_max_entries`.

liuchengxu:
Well, the ideas above are merely my quick thoughts, they don't have to be done in this PR or even later actually :(.

>What is the use case scenario of the yank timestamp?

I just thought it would be interesting to know about the time of that yank operation. It's essentially unimportant.

>getreg() in the above code executes before reading of viminfo/shada file before the VimEnter event.
Good to know! Then I'm ok with that.

autoload/clap/provider/yanks.vim
liuchengxu: Also add the doc about this option in doc/clap.txt.

plugin/clap.vim
```diff
@@ -19,3 +19,11 @@ augroup ClapBuffers
   autocmd BufWinEnter,WinEnter * let g:__clap_buffers[bufnr('')] = reltimefloat(reltime())
   autocmd BufDelete * silent! call remove(g:__clap_buffers, expand('<abuf>'))
 augroup END
+
+" yanks provider
+if get(g:, 'clap_enable_yanks_provider', 1)
```
liuchengxu: And add the doc about this option in doc/clap.txt.

liuchengxu:
Merged, thanks! Nice job @ratheesh  👍

PR #105, opened 2019-10-31
yanks: set minimum length to store in yanks stack

ratheesh:
Set the minimum length of yanked text to be stored in yanks stack. This
can be configured through a variable 'g:clap_provider_yanks_min_len'.

This feature is disabled by default(set to 1).

### DIFF
--- a/autoload/clap/provider/yanks.vim
+++ b/autoload/clap/provider/yanks.vim
@@ -6,11 +6,16 @@ let s:save_cpo = &cpoptions
 set cpoptions&vim
 
 let s:yank_history = []
-let s:yanks = {}
-let s:max_yanks = get(g:, 'clap_provider_yanks_max_entries', 20)
+let s:yanks        = {}
+let s:max_yanks    = get(g:, 'clap_provider_yanks_max_entries', 20)
+let s:min_len      = get(g:, 'clap_provider_yanks_min_len', 1)
 
 function! clap#provider#yanks#collect() abort
   let last_yanked = getreg('"')
+
+  if len(last_yanked) < s:min_len
+    return
+  endif
 
   if !empty(s:yank_history) && last_yanked == s:yank_history[0]
     return
--- a/doc/clap.txt
+++ b/doc/clap.txt
@@ -648,6 +648,13 @@ g:clap_provider_yanks_max_entries             *g:clap_provider_yanks_max_entries
 
   Set this variable to limit maximum yanks entries.
 
+g:clap_provider_yanks_min_len             *g:clap_provider_yanks_min_len*
+
+  Type: |Number|
+  Default: `1`
+
+  Set minimum length of yanked text to be stored in yanks stack. Setting this
+  variable to less than 2 has no effect.
 
 -------------------------------------------------------------------------------
 6.3 Other Provider Options                          *clap-other-provider-options*


autoload/clap/provider/yanks.vim
```diff
@@ -6,13 +6,14 @@ let s:save_cpo = &cpoptions
...
-let s:max_yanks = get(g:, 'clap_provider_yanks_max_entries', 20)
+let s:yanks        = {}
+let s:max_yanks    = get(g:, 'clap_provider_yanks_max_entries', 20)
+let s:min_len      = get(g:, 'clap_provider_yanks_min_len', 1)
 
 function! clap#provider#yanks#collect() abort
   let last_yanked = getreg('"')
 
-  if !empty(s:yank_history) && last_yanked == s:yank_history[0]
+  if !empty(s:yank_history) && last_yanked == s:yank_history[0] || len(last_yanked) < s:min_len
```
liuchengxu:
```suggestion
  if !empty(s:yank_history) && last_yanked == s:yank_history[0] && len(last_yanked) >= s:min_len
```
This is easier to understand IMO.

alerque:
Why are you keeping your own stack of yanks at all instead of just using the existing registers? I would have expected to see a 1-1 matchup between `:reg` and `:Clap yanks`. Especially on Neovim that persists these using *shada*...

liuchengxu:
I think you are actually talking about `:Clap registers` which is much more like `:reg`, but it has not been implemented.

`:Clap yanks` tracks the history of `"` register, but `:reg` keeps the last content of all kinds of register, that's the difference. I don't know much about neovim shada.

alerque:
Under what circumstances would you expect the history of the `"` register to be meaningfully different from the register stack already kept by vim? I could think of a lot of ways to make them differ artificially (yanking directly into specific other registers), but I can't think of any that would be meaningful to me when wanting to see my yank history. Why not combine `:Clap yanks` and `:Clap registers` ... unless somebody can define why the use case for them would be different.

ratheesh:
@alerque IMO, there is only one yank register(Register `0`). Registers `0` to `9` are used when text is deleted. This is why we need yank stack to keep track of yanks(which includes deleted text as well).
Correct me if I am wrong.

autoload/clap/provider/yanks.vim
ratheesh: This logic will not work as we need to return if `len(yanked text)` is less than `s:min_len`.
I tested and confirmed.

liuchengxu: Yeap, the code is not corrent directly, but what I want to express is not to use multiple `** && ** || **` if possible, which makes the priority not so clear in the long term. Then you can try this:

```vim
if len(last_yanked) < s:min_len
  return
endif
```

ratheesh: agreed and done.

alerque:
@Ratheesh I stand corrected. Apparently I hadn't had enough coffee yet this morning (GMT+3) because that is a very obvious difference that makes for significantly different usages.

I think you're wrong about there only being one yank register because you can yank to any register, but tracking the difference between things that end up in a register because they were yanked vs. because they were deleted is significant and completely justifies having both.

ratheesh:
> I think you're wrong about there only being one yank register because you can yank to any register, but tracking the difference between things that end up in a register because they were yanked vs. because they were deleted is significant and completely justifies having both.

Yes, when I meant one yank register, I meant default register when yanked without passing named register explicitly.

PR #107, opened 2019-10-31
Implement basic registers provider

ratheesh:
This commit adds support for basic registers provider. Preview feature
is currently not supported.

This implementation is loosely based on
https://github.com/mattn/ctrlp-register.git


liuchengxu:
What a coincidence! I have just implemented the registers provider and pushed to the master. Thanks for the PR anyway! You could review and improve that if you like :).

PR #133, opened 2019-11-19
Dont cd back if sink changes dir

meck:
I ran into a issue when fiddling with a provider for changing directories, where the enable_rooter function reverted the changes.


liuchengxu:
Thanks!

PR #143, opened 2019-12-01
Fix pressing C-J/K without input causes an error.

hokorobi:
image


liuchengxu:
Merged, thanks!

hokorobi:
Thak you.

PR #152, opened 2019-12-10
Add git_diff_files provider

kit494way:
Provider for files which is managed by git and have uncommitted changes.

### DIFF
--- a/autoload/clap/provider/git_diff_files.vim
+++ b/autoload/clap/provider/git_diff_files.vim
@@ -8,10 +8,17 @@ let s:git_diff_files = {}
 
 function! s:git_diff_files.source() abort
   if !executable('git')
-    return ['git executable not found']
+    call clap#helper#echo_error('git executable not found')
+    return []
   endif
 
-  return map(systemlist('git status -s -uno'), 'split(v:val)[-1]')
+  let changed = systemlist('git status -s -uno')
+  if v:shell_error
+    call clap#helper#echo_error('Error occurs on calling `git status -s -uno`, maybe you are not in a git repo.')
+    return []
+  else
+    return map(changed, 'split(v:val)[-1]')
+  endif
 endfunction
 
 let s:git_diff_files.sink = 'e'
--- a/autoload/clap/provider/git_diff_files.vim
+++ b/autoload/clap/provider/git_diff_files.vim
@@ -11,11 +11,7 @@ function! s:git_diff_files.source() abort
     return ['git executable not found']
   endif
 
-  redir => l:output
-    silent execute '!git status -s -uno'
-  redir END
-  let l:output = filter(split(l:output, '\n')[1:], 'len(trim(v:val))')
-  return map(l:output, 'split(v:val)[-1]')
+  return map(systemlist('git status -s -uno'), 'split(v:val)[-1]')
 endfunction
 
 let s:git_diff_files.sink = 'e'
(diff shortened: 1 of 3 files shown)


autoload/clap/provider/git_diff_files.vim
```diff
@@ -0,0 +1,27 @@
...
+let s:git_diff_files = {}
+
+function! s:git_diff_files.source() abort
+  if !executable('git')
+    return ['git executable not found']
+  endif
+
+  redir => l:output
+    silent execute '!git status -s -uno'
+  redir END
```
liuchengxu: It's simpler to use `system()`, `:h system()`.

kit494way:
Thanks for your advice.
I changed to use `systemlist()` and make it more simple.

autoload/clap/provider/git_diff_files.vim
```diff
@@ -0,0 +1,23 @@
...
+set cpoptions&vim
+
+let s:git_diff_files = {}
+
+function! s:git_diff_files.source() abort
+  if !executable('git')
+    return ['git executable not found']
+  endif
+
+  return map(systemlist('git status -s -uno'), 'split(v:val)[-1]')
```
liuchengxu: According to the help of `systemlist()`, it will return an empty string on error, then applying a map on it will raise an E712 error.

```suggestion
  let untracked = systemlist('git status -s -uno')
  if untracked ==# ''
     return ['Error occurs on calling `git status -s -uno`, maybe you are not in a git repo.']
  else
    return map(untracked, 'split(v:val)[-1]')
  endif
```

autoload/clap/provider/git_diff_files.vim
```diff
@@ -0,0 +1,23 @@
+" Author: KITAGAWA Yasutaka <kit494way@gmail.com>
+" Description: List the files which is managed by git and have uncommitted changes.
```
liuchengxu: From the help of `git status`, `-u[<mode>], --untracked-files[=<mode>]`, `git_untracked_files` is a better name than `git_diff_files`? What do you think?

autoload/clap/provider/git_diff_files.vim
kit494way: Thanks for your suggestion.
I missed error handling.
But `untracked ==# ''` causes an error when `systemlist()` succeeds.
So, based on the way of [commits provider](https://github.com/liuchengxu/vim-clap/blob/6045f807046a1dec301f7573ab2ead5e4fde705a/autoload/clap/provider/commits.vim#L9),  I changed to check `v:shell_error` and use `clap#helper#echo_error()` to report error.

autoload/clap/provider/git_diff_files.vim
kit494way: Option `-uno` means "Show no untracked files", so `git_diff_files` is better.
Behavior of the option `-u[<mode>]` depends on `mode`.
Default mode `all` shows untracked files, but mode `no` does not show untracked files.

autoload/clap/provider/git_diff_files.vim
liuchengxu: Actually it's ok to return the error message directly as they will be shown in the display window. But I'm also fine with just printing the error :(.

autoload/clap/provider/git_diff_files.vim
liuchengxu: Ok, I have no problem then. Can you please also update the README and help doc?

kit494way: Yes, I will.

kit494way: I updated the README and help doc.

liuchengxu:
Merged, thanks!

PR #157, opened 2019-12-12
Fix async external filter on typing falling back to sync

miguelbacalhau:
When using external async filters vim-clap will revert back to the sync internal filter unless the provider returns more than 10000 entries. 

From what I could tell internally what is happening is that vim-clap is storing some cache in memory and if there is cache it will always run synchronously. This results in losing fuzzy matching when using external filters(unless you have the internal one written in phyton).

With this PR if there is a external filter set then even if there is cache the filtering will be done async.

### DIFF
--- a/autoload/clap/impl.vim
+++ b/autoload/clap/impl.vim
@@ -240,7 +240,8 @@ endfunction
 "             on_move
 "
 function! clap#impl#on_typed() abort
-  if exists('g:__clap_forerunner_result')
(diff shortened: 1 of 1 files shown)


liuchengxu:
Merged, thanks! By the way, once #147 is merged, vim-clap will be able to handle 200,000+ items decently using the built-in fuzzy filter. If the rust extension is installed, then you don't have to use the external filter. I believe people will have more motivations to install Python and Rust :(.

PR #160, opened 2019-12-14
Change Vec::new() to Vec::with_capacity()

ImmemorConsultrixContrarie:
The required length of those vectors is the exact length of `ranked` vector. It is faster and takes less memory to allocate vector once with known capacity than automatic resize through `push()`.


liuchengxu:
Thanks! :beers:

PR #166, opened 2019-12-16
fix empty initial list for external filters

miguelbacalhau:
When initializing the window for external filters some providers can show results before filtering I find this way to be more clear and makes more sense to show results to be filtered if possible.

I am not sure if this is the best way to do it but it I tested and it seem to work for all commands

### DIFF
--- a/autoload/clap/api.vim
+++ b/autoload/clap/api.vim
@@ -516,8 +516,7 @@ function! s:init_provider() abort
   endfunction
 
   function! provider.init_display_win() abort
-    if self.is_pure_async()
-          \ || clap#api#has_externalfilter()
+    if !has_key(g:clap.provider._(), 'source')
       return
     elseif self.type == g:__t_string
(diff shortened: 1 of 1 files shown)


liuchengxu:
This change is essentially equivalent to removing `clap#api#has_externalfilter()`, but `clap#api#has_externalfilter()` is intentional. If you choose to use the external filter(async), then the built-in filter(sync) won't be used, for not everyone has installed the optional `python` dependency.

miguelbacalhau:
I also don't have the `python` dependency installed. Is there a better way to populate the list? Right now it will appear empty and only will show results on type. I the 'after' image below is what I am trying to accomplish.

Before:
Screen Shot 2019-12-17 at 11 27 22

After:
Screen Shot 2019-12-17 at 11 26 53

liuchengxu:
Please pull the latest commit to see if that is what you want.

miguelbacalhau:
Yes that is it thank you 👍

PR #169, opened 2019-12-17
Quickfix provider

kit494way:
Provider for quickfix list.

### DIFF
--- a/autoload/clap/provider/quickfix.vim
+++ b/autoload/clap/provider/quickfix.vim
@@ -1,0 +1,35 @@
+" Author: KITAGAWA Yasutaka <kit494way@gmail.com>
+" Description: List the entries of the quickfix list.
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+let s:quickfix = {}
+
+function! s:quickfix.source() abort
+  " Ignore quickfix list entries with non-existing buffer number.
+  let qflist = filter(getqflist(), 'v:val["bufnr"]')
+
+  return map(getqflist(), 's:to_grepformat(v:val)')
+endfunction
+
+function! s:to_grepformat(quickfix) abort
+  let path = bufname(a:quickfix['bufnr'])
+  let line_col = a:quickfix['lnum'].' col '.a:quickfix['col']
+  return path.'|'.line_col.'|'.a:quickfix['text']
+endfunction
+
+function! s:quickfix.sink(selected) abort
+  let [fpath, line_col] = split(a:selected, '|')[:1]
+  let [lnum, column] = split(line_col, ' col ')
+
+  execute 'edit' fpath
+  noautocmd call cursor(lnum, column)
+endfunction
+
+let s:quickfix.enable_rooter = v:true
+
+let g:clap#provider#quickfix# = s:quickfix
+
+let &cpoptions = s:save_cpo
+unlet s:save_cpo
--- a/autoload/clap/provider/quickfix.vim
+++ b/autoload/clap/provider/quickfix.vim
@@ -27,6 +27,15 @@ function! s:quickfix.sink(selected) abort
   noautocmd call cursor(lnum, column)
 endfunction
 
+function! s:quickfix.on_enter() abort
+  if has('nvim')
+    call g:clap.display.goto_win()
+    runtime! syntax/qf.vim
+    call g:clap.input.goto_win()
+  else
+    call win_execute(g:clap.display.winid, 'runtime! syntax/qf.vim')
+  endif
+endfunction
 
 let g:clap#provider#quickfix# = s:quickfix
 
--- a/autoload/clap/provider/quickfix.vim
+++ b/autoload/clap/provider/quickfix.vim
@@ -10,7 +10,7 @@ function! s:quickfix.source() abort
   " Ignore quickfix list entries with non-existing buffer number.
   let qflist = filter(getqflist(), 'v:val["bufnr"]')
 
-  return map(getqflist(), 's:to_grepformat(v:val)')
+  return map(qflist, 's:to_grepformat(v:val)')
 endfunction
 
 function! s:to_grepformat(quickfix) abort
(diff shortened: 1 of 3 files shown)


autoload/clap/provider/quickfix.vim
```diff
@@ -0,0 +1,35 @@
...
+" Description: List the entries of the quickfix list.
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+let s:quickfix = {}
+
+function! s:quickfix.source() abort
+  " Ignore quickfix list entries with non-existing buffer number.
+  let qflist = filter(getqflist(), 'v:val["bufnr"]')
```
liuchengxu: This is unused actually.

autoload/clap/provider/quickfix.vim
```diff
@@ -0,0 +1,35 @@
...
+
+function! s:quickfix.sink(selected) abort
+  let [fpath, line_col] = split(a:selected, '|')[:1]
+  let [lnum, column] = split(line_col, ' col ')
+
+  execute 'edit' fpath
+  noautocmd call cursor(lnum, column)
+endfunction
+
+let s:quickfix.enable_rooter = v:true
```
liuchengxu: `enable_rooter` is usually for these providers that need to run some `cwd` related commands. It should be unneccessary in this provider.

autoload/clap/provider/quickfix.vim
```diff
@@ -0,0 +1,35 @@
...
+
```
liuchengxu: Try adding `let s:quickfix.on_enter = { -> g:clap.display.setbufvar('&ft', 'qf')}` for getting the qf highlight.

autoload/clap/provider/quickfix.vim
kit494way: My mistake. Thanks for catching that.

autoload/clap/provider/quickfix.vim
kit494way: Thanks. I understand now.

autoload/clap/provider/quickfix.vim
kit494way: Thanks for the suggestion. I'll try this.

kit494way:
I tried `let s:quickfix.on_enter = { -> g:clap.display.setbufvar('&ft', 'qf')}`, but window was fixed to bottom like this.
スクリーンショット 2019-12-18 13 27 10

Is there any better filetype? Or is it ok to be unnset?

liuchengxu:
That window is not vim-clap window. It should like this, the display window of vim-clap has the same highlight of quickfix window:

屏幕快照 2019-12-18 下午1 12 51

Do you have some quickfix autocmds? If so, we could use some safer to enable the qf highlight. See if the following patch works for you:

```diff
diff --git a/autoload/clap/provider/quickfix.vim b/autoload/clap/provider/quickfix.vim
index d91bfec..6ed2344 100644
--- a/autoload/clap/provider/quickfix.vim
+++ b/autoload/clap/provider/quickfix.vim
@@ -27,6 +27,18 @@ function! s:quickfix.sink(selected) abort
   noautocmd call cursor(lnum, column)
 endfunction

+function! s:quickfix.on_enter() abort
+  if has('nvim')
+    call g:clap.display.goto_win()
+    runtime! syntax/qf.vim
+    call g:clap.input.goto_win()
+  else
+    call win_execute(g:clap.display.winid, 'runtime! syntax/qf.vim')
+  endif
+endfunction
```

The highlight is not critical, just for making the UI look better.

autoload/clap/provider/quickfix.vim
```diff
@@ -0,0 +1,34 @@
...
+  " Ignore quickfix list entries with non-existing buffer number.
+  let qflist = filter(getqflist(), 'v:val["bufnr"]')
+
+  return map(qflist, 's:to_grepformat(v:val)')
+endfunction
+
+function! s:to_grepformat(quickfix) abort
+  let path = bufname(a:quickfix['bufnr'])
+  let line_col = a:quickfix['lnum'].' col '.a:quickfix['col']
+  return path.'|'.line_col.'|'.a:quickfix['text']
```
liuchengxu: `text` is seemingly trimmed for the quickfix entry, I believe we'd better learn that too.

autoload/clap/provider/quickfix.vim
```diff
@@ -0,0 +1,34 @@
...
+let s:quickfix = {}
+
+function! s:quickfix.source() abort
+  " Ignore quickfix list entries with non-existing buffer number.
+  let qflist = filter(getqflist(), 'v:val["bufnr"]')
+
+  return map(qflist, 's:to_grepformat(v:val)')
+endfunction
+
+function! s:to_grepformat(quickfix) abort
```
liuchengxu: Is `qf_entry` a better name?

autoload/clap/provider/quickfix.vim
kit494way: I didn't realize that. Thanks.

> I believe we'd better learn that too.

I agree.

autoload/clap/provider/quickfix.vim
kit494way: Yes. I'll fix that.

kit494way:
I tried your [patch](https://github.com/liuchengxu/vim-clap/pull/169#issuecomment-566872613).
Then, vim-clap window is arranged to correct position.

2019-12-18_22h41_26

The highlight is different from quickfix window, but I found my color scheme is something wrong.
So your patch works good. Thank you so match.

liuchengxu:
Not sure what happened, but the colors does not look good in you case, you may want to checkout https://github.com/liuchengxu/vim-clap#change-highlights for the better looking :(.

kit494way:
> `text` is seemingly trimmed when the quickfix entry is displayed in the quickfix window

Although I couldn't find any description about the format of text in the quickfix window in vim help,
I found the code which trims text in (neo)vim source code.

- https://github.com/neovim/neovim/blob/v0.4.3/src/nvim/quickfix.c#L3393-L3397
- https://github.com/vim/vim/blob/v8.2.0023/src/quickfix.c#L4475-L4479

So, to use the same format with (neo)vim, I changed my code to use like this
- Remove leading white spaces
- Put a space to head of text
- Replace newline and following spaces with a space

liuchengxu:
Merged, thanks!

PR #180, opened 2019-12-25
Move fzy algo from rff crate into local file and some other small tweaks to Rust code

ImmemorConsultrixContrarie:
Original (MIT licensed) "rff" crate has `terminal` module which utilizes `std::os::unix` thus it doesn't compile on non-unix OS, and could not be "cargo check"ed and tested.
If it is not accepted, all tweaks that don't depend on moving fzy algorithm are done in "Less allocations, more laziness" commit.


pythonx/clap/extracted_fzy/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
+!.cargo
+*.lock
```
liuchengxu: I don't think the [faq](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) applies to our situation. Although it's a library, it's unpublished and actually used for vim-clap exclusively. No others depend on it, and we need to ensure people are always able to compile this Rust extension on their own. Keeping `Cargo.lock` is better for our case.

liuchengxu:
Overall looks good to me, except ignoring the `Cargo.lock`  I'm not against moving the fzy algo impl into vim-clap if that helps non-unix users. We can use a dedicated fzy library once https://github.com/stewart/rff/issues/2 is resolved, but rff is seemingly not much active now :(.

ImmemorConsultrixContrarie:
Okay, will revert gitignoring .lock files and add my .locks to the PR, once I'm back at home.

ImmemorConsultrixContrarie:
Ready to merge now, I guess?
@liuchengxu

liuchengxu:
Merged, thanks!

PR #184, opened 2019-12-28
Add forerunner job status sign and a delay timer for running maple

### LINKED ISSUE #140: Performance profile
### Introduction

The one selling point of vim-clap is in pure vimscript, having a good platform compatibility and minimal pain to set it up, which is apparently an advantage. However, it becomes a disadvantage when you do **a large scale searching/filtering** job for the limit of vimscript language.

### Profile of async job at scale

For instance, using `:Clap files` from my home directory and input `sr`, the result is `[791970/823172]`. The following profile info is based on this commit dfffb5e.

Profile of NeoVim:

```
FUNCTIONS SORTED ON TOTAL TIME
count  total (s)   self (s)  function
16737   2.416528   0.592996  <SNR>181_on_event()
16732   1.816953   0.267504  <SNR>181_append_output()
16731   1.505205   0.146196  <SNR>181_set_matches_count()
16732   1.379571   0.310434  clap#impl#refresh_matches_count()
16732   0.821710   0.141712  clap#indicator#set_matches()
16732   0.679998   0.317742  <SNR>183_apply_indicator()
    3   0.492807   0.472198  <SNR>172_on_event()
16732   0.362257   0.177423  <SNR>183_padding()
16733   0.247077   0.140285  clap#sign#reset_to_first_line()
    8   0.195782             provider#python3#Call()
16740   0.184996   0.100084  16()
    5   0.147412             UltiSnips#TrackChange()
16732   0.106776             <SNR>184_sign_of_first_line()
16740   0.084913             clap#util#nvim_buf_get_first_line()
    3   0.078762   0.000110  clap#handler#on_typed()
    2   0.078562   0.003062  41()
    2   0.075139   0.000245  clap#impl#on_typed()
    5   0.067303   0.000438  clap#rooter#run()
    2   0.063340   0.004730  <SNR>179_on_typed_async_impl()
    2   0.052697   0.000056  clap#dispatcher#job_start()
```

Profile of Vim:

```
FUNCTIONS SORTED ON TOTAL TIME
count  total (s)   self (s)  function
1613221  70.847359  38.246657  <SNR>172_out_cb()
1613224  31.405335  18.249628  clap#job#vim8_job_id_of()
1613229  13.155912             clap#job#parse_vim8_job_id()
791899   1.191679             <SNR>172_handle_cache()
    1   0.537275   0.517298  <SNR>167_close_cb()
   61   0.183191   0.007911  ElelineGitBranch()
   60   0.170001   0.154896  FugitiveExtractGitDir()
   18   0.021677   0.001292  <SNR>162_popup_filter()
    2   0.020385   0.000444  142()
    3   0.020079   0.019061  clap#impl#refresh_matches_count()
    1   0.019937   0.000054  <SNR>167_on_complete()
    2   0.019428   0.000085  <SNR>162_apply_input()
    2   0.019320   0.001760  117()
    5   0.018673   0.001056  clap#rooter#run()
    2   0.016882   0.000210  clap#impl#on_typed()
   10   0.014478   0.003716  traces#init()
    2   0.014044   0.000092  <SNR>172_post_check()
    2   0.013604   0.000199  <SNR>172_on_exit_common()
    2   0.013177   0.000034  clap#impl#add_highlight_for_fuzzy_indices()
    2   0.013143   0.010832  <SNR>170_apply_add_highlight()
```

For the result set having about 1M items, neovim is still tolerable if you seldom run into that case, but for vim it's totally unusable. The core idea of vim-clap is to spawn a new job on your typing and then react on the callback, the callback of the job is called too many times on the surface.

### Related providers

The performance issue only happens to these providers that can have a huge result set.

#### Searching

- `:Clap files`
- `:Clap grep`

#### Filtering

The current strategy is to use **the built-in fzy written in python running synchronously** when the items are no more than 10,000, switching the async way if it's more than the threshold and then it meets the same callback issue.

- `:Clap blines`
- `:Clap lines`
- `:Clap bcommits`
- `:Clap commits`

The challenge is to make `clap#filter#()` fast enough even when `candidates` is terrific.

https://github.com/liuchengxu/vim-clap/blob/77631c150927650eeaa4067f8a1889d4578aea5e/autoload/clap/filter.vim#L164

### Possible solution

- [ ] `python` multi thread. Since we have introduced the optional dependency for the built-in fzy support, we could also use the thread of python to improve the performance as suggested in https://github.com/liuchengxu/vim-clap/issues/75#issuecomment-544767636. As I'm not a python expert, I'm not sure if this will land in vim-clap eventually.

- [ ] `lua`. Could be faster than vimscript at least in some filtering cases. I observe something promissing based on the [this script measuring the vimscript with the other script languages](https://groups.google.com/d/msg/vim_dev/__gARXMigYE/orGEJs3PBwAJ):

    - `lua` is 4x faster than `python`.

    - `luajit` is **100x** faster than `python`.

- [x] Boost python filtering using Rust. I'm interested if https://github.com/rochacbruno/rust-python-example would help, which can replace `fuzzy_match()`: https://github.com/liuchengxu/vim-clap/blob/f4d280caa8e5e663d783740d4209d5200efbbc7d/pythonx/clap/fzy.py#L9 The result shows that using Rust could be 10x faster, see #146, #147.
        
- `job`
   - [-] ~Use a timer for reading the output of vim job periodly, then the callback won't be called that many times~. [This branch](https://github.com/liuchengxu/vim-clap/tree/timer-for-vim-job) is my try with timer. Using a timer helps a lot, but some messages are dropped unexpected, not sure if that's bug of Vim or my code. I did not investigate further as I have achived the same goal with `maple` as follows.
   - [x] Don't have to output all the fuzzy filtered items, we only show the top N candidates and care about the number of total candidates. The payload of async job will be significantly decreased if only these _essential_ information is returned.(#181)

liuchengxu:
- Add the forerunner job status sign:
  - `!files`: the forerunner job is not finished yet. If you type something rightnow, the result is incomplete and uncertain.

  - `*files`: the forerunner job is done, now the responose of vim-clap should be fast and deterministic.
 
  The forerunner job is normally super fast and imperceptible unless you have 1 million items, in which case it may take seconds but Vim/NeoVim is still well responsive and you won't feel any typing lag.

- Add the delay for running maple.

clap-fix-perf

I proposed serveral solutions for the perfornce improvement in #140. Although I have not tried all of them, the `maple` extension written in Rust already works very well. If you want to a performant vim-clap, please [install Rust](https://www.rust-lang.org/tools/install) and then run `:call clap#helper#build_all()` in Vim/NeoVim.

Close #140.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,9 @@ Vim-clap is a modern generic interactive finder and dispatcher, based on the new
 * [Requirement](#requirement)
   * [Optional](#optional)
     * [`python`](#python)
-    * [`maple`](#maple)
-    * [Rust extension](#rust-extension)
+    * [`Rust`](#rust)
+      * [`maple` binary](#maple-binary)
+      * [Python dynamic module](#python-dynamic-module)
 * [Installation](#installation)
 * [Usage](#usage)
   * [Commands](#commands)
@@ -85,30 +86,41 @@ Vim-clap is a modern generic interactive finder and dispatcher, based on the new
   $ python3 -m pip install pynvim
   ```
 
-#### `maple`
+#### `Rust`
 
-`maple` is essentially a tiny wrapper of [skim](https://github.com/lotabout/skim) and [fzy](https://github.com/jhawthorn/fzy), with the matched indices exposed to be highlighted in vim-clap's async providers.
+If you have installed Rust on your system, specifically, `cargo` executable exists, you can build the extra tools for a performant and nicer vim-clap using one single command `:call clap#helper#build_all()`.
 
-Use `:call clap#helper#build_maple()` or install maple manually:
+##### `maple` binary
 
-```bash
-# Compile the release build
-cargo build --release
+`maple` mainly serves two functions:
 
-# Or use cargo install
-cargo install --path . --force
-```
+1. Expose the fuzzy matched indices so that the matched elements can be highlighted in vim-clap, being a tiny wrapper of external fuzzy filter [fzf](https://github.com/junegunn/fzf) and [fzy](https://github.com/jhawthorn/fzy). Once you installed `maple`, fzy/skim binary are unneeded as `maple` does not rely the binary directly but reuses their filter algorithm internally.
 
-#### Rust extension
+2. Reduce the overhead of async job of Vim/NeoVim dramastically.
 
-Use `:call clap#helper#build_rust_ext()` to install the Rust extension for 10x faster fuzzy filter than the Python one.
+To install `maple` you can use the helper function and run `:call clap#helper#build_maple()`, or install it manually:
+
+  ```bash
+  # Compile the release build
+  cargo build --release
+
+  # Or use cargo install globally
+  cargo install --path . --force
+  ```
+
+##### Python dynamic module
+
+Use `:call clap#helper#build_rust_ext()` to install the Python dynamic module written in Rust for 10x faster fuzzy filter than the Python one. Refer to the post [Make Vim Python plugin 10x faster using Rust](http://liuchengxu.org/posts/speed-up-vim-python-plugin-using-rust/) for the whole story.
 
 ## Installation
+
+Using [vim-plug](https://github.com/junegunn/vim-plug):
 
 ```vim
 Plug 'liuchengxu/vim-clap'
 
-" Build the all optional dependency, cargo is needed.
+" The do hook is highly recommended.
+" It will try to build all the optional dependency if cargo exists on your system.
 Plug 'liuchengxu/vim-clap', { 'do': function('clap#helper#build_all') }
 ```
 
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -40,6 +40,9 @@ let s:provider_alias = {
 let s:provider_alias = extend(s:provider_alias, get(g:, 'clap_provider_alias', {}))
 let g:clap#provider_alias = s:provider_alias
 
+let g:clap_forerunner_status_sign_done = get(g:, 'clap_forerunner_status_sign_done', '*')
+let g:clap_forerunner_status_sign_running = get(g:, 'clap_forerunner_status_sign_running', '!')
+
 let g:clap_no_matches_msg = get(g:, 'clap_no_matches_msg', 'NO MATCHES FOUND')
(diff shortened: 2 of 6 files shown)


ImmemorConsultrixContrarie:
I should say, all this flickering looks just bad. Maybe text field should be cleared by new results, not just wiped at the time user changes input? Because [that](https://user-images.githubusercontent.com/8850248/67620599-3b1c9a80-f83b-11e9-8d8c-72bfae9d9177.gif) doesn't look as bad as gif in this PR.

Also, this may be more perceptible and smooth if results would be shown line by line with very short delay, not just dropped on the screen all at once. Though I'm not quite sure about it.

liuchengxu:
The `redraw` is caused by the `job` inherently to my knowledge, not the rendering strategy used by vim-clap. Rendering the content linewise does not help.

`:Clap blines` in this [gif](https://user-images.githubusercontent.com/8850248/67620599-3b1c9a80-f83b-11e9-8d8c-72bfae9d9177.gif) does not use the async `job`, but the built-in filter which is sync, so it does not have the `redraw` issue.

liuchengxu:
Hmm, I just tried not clearing the previous content anyway, which seems to help indeed. Can you also check it out, see if it's really helpful? @ImmemorConsultrixContrarie 

```diff
diff --git a/autoload/clap/impl.vim b/autoload/clap/impl.vim
index ec5c2cd..092a389 100644
--- a/autoload/clap/impl.vim
+++ b/autoload/clap/impl.vim
@@ -201,7 +201,7 @@ function! s:on_typed_async_impl() abort
     return
   endif
 
-  call g:clap.display.clear()
+  " call g:clap.display.clear()
 
   let cmd = g:clap.provider.source_async_or_default()
```

ImmemorConsultrixContrarie:
> Hmm, I just tried not clearing the previous content anyway, which seems to help indeed. Can you also check it out, see if it's really helpful?

Err... No. I don't use either vim or vim-clap. I just write code blindly, heh.

liuchengxu:
> I just write code blindly

I actually very admire this..., I can't live without these tools.

Anyway I have confirmed that not clearing the outdated result immedidately indeed avoids the visual flicker. Thank you! @ImmemorConsultrixContrarie

PR #188, opened 2019-12-30
Lint PRs using Reviewdog

alerque:
Since using Vint in Github Actions I've run across this action somebody setup to run Vint inside or Reviewdog. The end result should be the same, but instead of mucking around in the console output of a CI run to find linting errors, a bot will post PR review comments on the actual line number in question. This should make it easier on contributors.

In order to use this I had to drop the hack that was skipping the heredoc into Python. Predictably this turned up one linting error in that file too.


liuchengxu:
This looks great. Thanks! 👍

PR #192, opened 2020-01-01
Fix printf too many args error if maple exec not found

mrinterweb:
I was getting the error: "too many arguments to printf"
when the maple command was not available. This change works
for me.


liuchengxu:
Let me know if this doesn't work for you.

PR #198, opened 2020-01-02
Uppercase the NONE term attribute

runar:
A capitalised term attribute (None) results in a `E254: Cannot allocate color None` error message when running `:Clap files`. This pull request fixes that issue by uppercasing the attribute.


liuchengxu:
Thanks!

PR #215, opened 2020-01-05
Add working git_files based on git_diff_files and files providers

wbthomason:
The `git_files` provider does not work for me; I always get an empty window with no results. As well, the existing provider does not support open actions, etc.

This PR replaces the existing `git_files` provider with one constructed from code from the existing `git_diff_files` and `files` providers (both of which do work for me). It adds better error reporting, syntax highlighting like the `files` provider,  and handling of open actions.

I looked into getting this provider to support icons, like the files provider, but wasn't sure how to get that working.

### DIFF
--- a/autoload/clap/provider/git_files.vim
+++ b/autoload/clap/provider/git_files.vim
@@ -6,14 +6,32 @@ set cpoptions&vim
 
 let s:git_files = {}
 
-if executable('git')
-  let s:git_files.source = 'git ls-files '.(has('win32') ? '' : ' | uniq')
-else
-  let s:git_files.source = ['git executable not found']
-endif
+function! s:git_files.source() abort
+  if !executable('git')
+    call clap#helper#echo_error('git executable not found')
+    return []
+  endif
 
-let s:git_files.sink = 'e'
+  let changed = systemlist('git ls-files '.(has('win32') ? '' : ' | uniq'))
+  if v:shell_error
+    call clap#helper#echo_error('Error occurs on calling `git ls-files`. Maybe you are not in a git repo?')
+    return []
+  else
+    return map(changed, 'split(v:val)[-1]')
+  endif
+endfunction
+
+function! s:git_files.sink(selected) abort
+  if has_key(g:clap, 'open_action')
+    execute g:clap.open_action a:selected
+  else
+    execute 'edit' a:selected
+  endif
+endfunction
+
 let s:git_files.enable_rooter = v:true
+let s:git_files.support_open_action = v:true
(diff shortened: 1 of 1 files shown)


autoload/clap/provider/git_files.vim
```diff
@@ -6,14 +6,32 @@ set cpoptions&vim
...
-else
-  let s:git_files.source = ['git executable not found']
-endif
+function! s:git_files.source() abort
+  if !executable('git')
+    call clap#helper#echo_error('git executable not found')
+    return []
+  endif
+
+  let changed = systemlist('git ls-files '.(has('win32') ? '' : ' | uniq'))
```
liuchengxu: Simply changing the `source` to a function returning a List will make this provider always run in sync way. It's ok in most cases, but it will be a problem when the git files are too many, this system call could be slow, making vim-clap sluggish and you'll complain that.

Check out `converter` for adding an icon using VimL, https://github.com/liuchengxu/vim-clap#create-pure-async-provider.

I'll see why `git_files` returns nothing.

wbthomason: Ok. Would you suggest that a pure async provider would be better, then?

wbthomason: Hmm. I would guess not, since pure async providers don't seem to have a way to specify a source to run without a query (i.e. when `on_typed` doesn't really make sense, as here).

liuchengxu: Actually the non-pure async provider can also have the `converter` property, I forget to document that. But it's better to provide the icon support in the external Rust extension `maple` since we already use it.

wbthomason: Ok. I looked into that briefly, by modifying https://github.com/liuchengxu/vim-clap/blob/master/autoload/clap/maple.vim#L124 to also test against `git_files`, but it didn't add icons when I ran it. It seems that there are a number of places in the code with special-case handling for the providers that support icons; maybe it would make sense to refactor `maple` icon support to be a property of a provider, then just test if this property is set?

liuchengxu: Adding an icon support needs both `source` and `sink` to be tweaked. If there is a clear pattern, we could make it easier by introducing some property into the provider.

liuchengxu:
@wbthomason Can you check if #216 fixes the issue of returning nothing when using `git_files`?

wbthomason:
@liuchengxu It does, thank you! I will close this PR, as it now seems unnecessary. Thanks!

PR #222, opened 2020-01-09
Install the prebuilt binary automatically

### LINKED ISSUE #221: [Testers wanted] Release the prebuilt binary via GitHub Action
Since [v0.4](https://github.com/liuchengxu/vim-clap/releases/tag/v0.4), the prebuilt binary `maple` has been provided via the GitHub Action service. If you want to use the prebuilt binary:

### Manual install

1. download the binary at https://github.com/liuchengxu/vim-clap/releases/tag/v0.4 accordingly.
2. rename to:
    - windows: `maple.exe`
    - unix: `maple`
3. move `maple`/`maple.exe` to `bin` directory.

### Download the binary using the provided script

**Testers are wanted, especially for Windows:**

1. Switch to the [auto-install branch](https://github.com/liuchengxu/vim-clap/tree/auto-install).
2. If you have installed `cargo`:
    1. Run `cargo clean` outside Vim/NeoVim to clean the current local build if you have built the binary `maple`.
    2. Call `:call clap#helper#download_binary()` from Vim/NeoVim.
3. If you don't have `cargo` installed, call `:Clap install-binary!` from Vim/NeoVim.
4. Reopen Vim/NeoVim to see if it works as expected.

liuchengxu:
Close #221

### DIFF
--- a/autoload/clap/helper.vim
+++ b/autoload/clap/helper.vim
@@ -160,5 +160,19 @@ function! clap#helper#build_all(...) abort
   endif
 endfunction
 
+function! clap#helper#download_binary() abort
+  let cwd = fnamemodify(g:clap#autoload_dir, ':h')
+  let cmd = './install.sh'
+  call s:run_term(cmd, cwd, 'download the prebuilt maple binary successfully')
+endfunction
+
+function! clap#helper#install() abort
+  if executable('cargo')
+    call clap#helper#build_all()
+  else
+    call clap#helper#download_binary()
+  endif
+endfunction
+
 let &cpoptions = s:save_cpo
 unlet s:save_cpo
--- a/autoload/clap/helper.vim
+++ b/autoload/clap/helper.vim
@@ -160,10 +160,13 @@ function! clap#helper#build_all(...) abort
   endif
 endfunction
 
-" TODO: windows support
 function! clap#helper#download_binary() abort
   let cwd = fnamemodify(g:clap#autoload_dir, ':h')
-  let cmd = './install.sh'
+  if has('win32')
+    let cmd = 'Powershell.exe -File '.cwd.'\install.ps1'
+  else
+    let cmd = './install.sh'
+  endif
   call s:run_term(cmd, cwd, 'download the prebuilt maple binary successfully')
 endfunction
 
--- a/autoload/clap/helper.vim
+++ b/autoload/clap/helper.vim
@@ -166,7 +166,7 @@ function! clap#helper#download_binary() abort
   call s:run_term(cmd, cwd, 'download the prebuilt maple binary successfully')
 endfunction
 
-function! clap#helper#install() abort
+function! clap#helper#install(...) abort
   if executable('cargo')
     call clap#helper#build_all()
   else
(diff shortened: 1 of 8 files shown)


alerque:
This should be combined with variations on the plugin manage install commands, one each for how to install without the helper, one for a hook that builds the helper locally (but automatically in install for supported plugin managers), and one for how to fetch the pre-built.

Nice work by the way, still loving Clap!

autoload/clap/helper.vim
```diff
@@ -160,5 +160,19 @@ function! clap#helper#build_all(...) abort
...
+  let cwd = fnamemodify(g:clap#autoload_dir, ':h')
+  let cmd = './install.sh'
+  call s:run_term(cmd, cwd, 'download the prebuilt maple binary successfully')
+endfunction
+
+function! clap#helper#install() abort
+  if executable('cargo')
+    call clap#helper#build_all()
+  else
+    call clap#helper#download_binary()
```
alerque: Please don't make this download happen automatically unless _explicitly_ requested. I want my plugins to auto install but I don't want them downloading binary blobs in the process. Most of my systems have `cargo`, but on those that don't in particular I would rather the install fail than download a binary blob. This should require an explicit setup (such as an extra argument to `clap#helper#install(allow_binary)` that opts-in so that it would not _ever_ get accidentally triggered.

liuchengxu: You can easiy achive that by :

```vim
if executable('cargo')
  Plug 'liuchengxu/vim-clap', { 'do': function('clap#helper#install') }
else
  Plug 'liuchengxu/vim-clap'
endif
```

alerque: This is still a dangerous and brittle way. The problem for me is that you're forcing an opt-out model rather than an opt-in. As long as my RC file had an `if X` statement that exactly matched the one in the plugin it would be okay, but if you ever were to add an extra check (say a version check) in the install function it could end up downloading the blob again.

Please reverse the logic so that there is a opt-in mechanism. A single truthy argument to install could be used to enable the download. Once the code in introduced it will be much harder to reverse the action, if you just have people start out by using `clap#helper#install(1)` out of the gate it will be easy to keep this as an opt-in.

liuchengxu: `clap#helper#install(1)` is not allowed for vim-plug, but a command is possible, I'll have a look later.

liuchengxu: My quick thought is to use a command for the `do` hook:

```vim
" Try building the binary if cargo exists
Plug 'liuchengxu/vim-clap', { 'do': ':Clap install-binary' }

" Build the binary if cargo exists, otherwise try downloading the prebuilt binary.
Plug 'liuchengxu/vim-clap', { 'do': ':Clap install-binary!' }
```

alerque: That looks okay to me. As long as only the bang version ever allows the download and I don't include the bang version in my RC there isn't a danger that it will accidentally get downloaded and run. These could be further clarified by renaming them `build-binary` and `build-or-fetch-binary` so it was more clear what did what without looking up docs or the source.

PR #248, opened 2020-01-18
Add help tags provider

markwu:
Hi @liuchengxu,

I add a provider to list help tags.

I implement cache for better performance, but it seems no file cache folder in Clap, therefore, I use `fnamemodify($MYVIMRC, ':p:h')` as my default cache folder.

If you have better place, just let me know.

And, I have no idea how to implement preview,  so, no preview at this moment.

### DIFF
--- a/autoload/clap/cache.vim
+++ b/autoload/clap/cache.vim
@@ -1,0 +1,44 @@
+" Author: Mark Wu <markplace@gmail.com>
+" Description: Cache API for clap.
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+let s:path_separator = has('win32') ? '\' : '/'
+let s:clap_cache_directory = get(g:, 'clap_cache_directory', '')
+
+function! clap#cache#directory() abort
+  if empty(s:clap_cache_directory)
+    if has('nvim')
+      let user_cache = stdpath('cache')
+    elseif exists('$XDG_CACHE_HOME')
+      let user_cache = $XDG_CACHE_HOME
+    else
+      let user_cache = $HOME . s:path_separator . '.cache'
+    endif
+    let s:clap_cache_directory = user_cache . s:path_separator . 'clap'
+  endif
+
+  if !isdirectory(s:clap_cache_directory)
+    call mkdir(s:clap_cache_directory, 'p')
+  endif
+
+  return s:clap_cache_directory
+endf
+
+function! clap#cache#location_for(provider_id, fname) abort
+  if empty(a:provider_id)
+    call clap#helper#echo_error('provider_id cannnot be empty.')
+  endif
+
+  let provider_cache_directory = clap#cache#directory() . s:path_separator . a:provider_id
+
+  if !isdirectory(provider_cache_directory)
+    call mkdir(provider_cache_directory, 'p')
+  endif
+
+  return provider_cache_directory . s:path_separator . a:fname
+endfunction
+
+let &cpoptions = s:save_cpo
+unlet s:save_cpo
--- a/autoload/clap/provider/help_tags.vim
+++ b/autoload/clap/provider/help_tags.vim
@@ -1,24 +1,25 @@
 " Author: Mark Wu <markplace@gmail.com>
-" Description: List the help tags.
+" Description: List the help tags, ported from https://github.com/zeero/vim-ctrlp-help
 
 let s:save_cpo = &cpoptions
 set cpoptions&vim
 
-" Help tags support, ported from https://github.com/zeero/vim-ctrlp-help
-let s:help_tags_filecache = get(g:, 'clap_provider_help_tags_cache', fnamemodify($MYVIMRC, ':p:h') . '/clap_provider_help_tags_cache')
-let s:help_tags_memcache = []
+let s:help_tags_memory_cache = []
 
 function! s:help_tags_source() abort
-  if empty(s:help_tags_memcache)
-    if getftime(s:help_tags_filecache) > max(map(s:get_tags_files(), 'getftime(v:val)'))
-      let s:help_tags_memcache = readfile(s:help_tags_filecache)
+  let help_tags_cache_file = clap#cache#location_for('help_tags', 'help_tags.txt')
+  if empty(s:help_tags_memory_cache)
+    if getftime(help_tags_cache_file) > max(map(s:get_tags_files(), 'getftime(v:val)'))
+      if filereadable(help_tags_cache_file)
+        let s:help_tags_memory_cache = readfile(help_tags_cache_file)
+      endif
     else
-      let s:help_tags_memcache = s:get_tags_list()
-      silent! call writefile(s:help_tags_memcache, s:help_tags_filecache)
+      let s:help_tags_memory_cache = s:get_tags_list()
+      silent! call writefile(s:help_tags_memory_cache, help_tags_cache_file)
     endif
   endif
 
-  return s:help_tags_memcache
+  return s:help_tags_memory_cache
 endfunction
 
 function! s:get_tags_list() abort
--- a/autoload/clap/provider/help_tags.vim
+++ b/autoload/clap/provider/help_tags.vim
@@ -1,0 +1,66 @@
+" Author: Mark Wu <markplace@gmail.com>
+" Description: List the help tags.
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+" Help tags support, ported from https://github.com/zeero/vim-ctrlp-help
+let s:help_tags_filecache = get(g:, 'clap_provider_help_tags_cache', fnamemodify($MYVIMRC, ':p:h') . '/clap_provider_help_tags_cache')
+let s:help_tags_memcache = []
+
+function! s:help_tags_source() abort
+  if empty(s:help_tags_memcache)
+    if getftime(s:help_tags_filecache) > max(map(s:get_tags_files(), 'getftime(v:val)'))
+      let s:help_tags_memcache = readfile(s:help_tags_filecache)
+    else
+      let s:help_tags_memcache = s:get_tags_list()
+      silent! call writefile(s:help_tags_memcache, s:help_tags_filecache)
+    endif
+  endif
+
+  return s:help_tags_memcache
+endfunction
+
+function! s:get_tags_list() abort
+  let tagsfiles = s:get_tags_files()
+
+  let input_dict = {}
+  for tagsfile in tagsfiles
+    for line in readfile(tagsfile)
+      let items = split(line, "\t")
+      let tag_subject = items[0]
+      if !has_key(input_dict, tag_subject)
+        let input_dict[tag_subject] = printf("%-60s\t%s", tag_subject, items[1])
+      endif
+    endfor
+  endfor
+  let input = sort(values(input_dict))
+
+  return input
+endfunction
+
+function! s:get_tags_files() abort
+  let tagspaths = map(filter(split(&helplang, ','), 'v:val !=? "en"'), '"/doc/tags-".v:val')
+  call add(tagspaths, '/doc/tags')
+
+  let tagsfiles = []
+  for tagspath in tagspaths
+    call extend(tagsfiles, filter(map(split(&runtimepath, ','), 'v:val . tagspath'), 'filereadable(v:val)'))
+  endfor
+
+  return tagsfiles
+endfunction
+
+function! s:help_tags_sink(line) abort
+  let tag = get(split(a:line, "\t"), 0)
+  execute 'help' tag
+endfunction
+
+let s:help_tags = {}
+let s:help_tags.sink = function('s:help_tags_sink')
+let s:help_tags.source = function('s:help_tags_source')
+
+let g:clap#provider#help_tags# = s:help_tags
+
+let &cpoptions = s:save_cpo
+unlet s:save_cpo
--- a/doc/clap.txt
+++ b/doc/clap.txt
@@ -213,6 +213,25 @@ g:clap_provider_alias                                    *g:clap_provider_alias*
   provider name.
 
 
+g:clap_cache_directory                                  *g:clap_cache_directory*
+
+  Type: |String|
+  Default: `Undefined`
+
+  This variable controls the cache directory of clap. In default, clap will set
+  cache directory according to the following order:
+
+  1. If has('nvim'), the default cache directory is `stdpath('cache')/clap`
+  2. If has `$XDG_CACHE_HOME`, the default cache directory is `$XDG_CACHE_HOME/clap`
+  3. If not all above, the default cache directory is `$HOME/.cache/clap`
+
+  If you want to change it to your cache directory, just set this variable. For
+  example:
+>
+  let g:clap_cache_directory = $HOME . '/.vim/cache/clap'
+<
+
+
 g:clap_layout                                                    *g:clap_layout*
 
   Type: |Dict|
--- a/doc/clap.txt
+++ b/doc/clap.txt
@@ -145,6 +145,10 @@ to work everywhere out of the box, with fast response.
                          use `Clap grep ..` to grep from the parent directory.
 
 
+                                                     *:Clap-help_tags*
+:Clap help_tags          List the help tags.
+
+
                                                       *:Clap-history*
 :Clap history            List the open buffers and |v:oldfiles|.
 


markwu:
That's why use cache. If you still worry about read/write operations, maybe I can 

1. add expire time to avoid read tags files again to compare the file time
2. or maybe just gather tags file only once when provider startup, 
3. or maybe use memory cache, like `g:clap_memory_cache['help_tags'] = {'time': ..., 'tags': [....]}`. with above 2 ...

I test the current provider with `maple`, `fzy` and `fzf`, the performance seems okay for me.

markwu:
I just add two new changes:

1. Use memory cache to reduce read/write operations
2. Allow user to change cache file through `g:clap_help_tags_cache`

autoload/clap/provider/help_tags.vim
```diff
@@ -0,0 +1,70 @@
...
+    return input
+endfunction
+
+function! s:get_tags_files() abort
+    let tagspaths = []
+    for lang in split(&helplang, ',')
+        if 'en' !=? lang
+            call add(tagspaths, '/doc/tags-' . lang)
+        endif
+    endfor
```
liuchengxu: This for-loop can be simplified as:

```vim
let tagspaths = map(filter(split(&helplang, ','), 'v:val !=? "en"'), '"/doc/tags-".v:val')
```

markwu: Done!

autoload/clap/provider/help_tags.vim
```diff
@@ -0,0 +1,65 @@
...
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+" Help tags support, ported from https://github.com/zeero/vim-ctrlp-help
+let s:help_tags_filecache = get(g:, 'clap_provider_help_tags_cache', fnamemodify($MYVIMRC, ':p:h') . '/clap_provider_help_tags_cache')
+let s:help_tags_memcache = []
+
+function! s:help_tags_source() abort
+    if(empty(s:help_tags_memcache))
```
liuchengxu:
```suggestion
    if empty(s:help_tags_memcache)
```

markwu:
I already change indent to 2. if everything okay, I will squash them to 1 commit.

autoload/clap/provider/help_tags.vim
markwu: Ha, I forgot I am writing VIM not PHP. Done.

autoload/clap/provider/help_tags.vim
```diff
@@ -0,0 +1,65 @@
...
+
+function! s:get_tags_list() abort
+  let tagsfiles = s:get_tags_files()
+
+  let input_dict = {}
+  for tagsfile in tagsfiles
+    for line in readfile(tagsfile)
+      let tag = split(line, "\t")
+      let line_dict = {matchstr(line, '^[^\t]\+') : printf("%-60s\t%s", tag[0], tag[1])}
+      call extend(input_dict, line_dict, 'keep')
```
liuchengxu: This should be more readable:

```vim
let tag_subject = matchstr(line, '^[^\t]\+')
if !has_key(input_dict, tag_subject)
  let input_dict[tag_subject] = printf("%-60s\t%s", tag[0], tag[1])
endif
```

liuchengxu:
When you type something and then clear the whole input, the matches count seem to be unexpected and the result list is what I see on startup.

On startup:

屏幕快照 2020-01-18 下午11 39 31

Type something and then clear the input:

屏幕快照 2020-01-18 下午11 39 45

markwu:
@liuchengxu I saw this behavior, too. I suppose it is expected behavior. 

Because, I saw it happened in `Clap grep`, too

autoload/clap/provider/help_tags.vim
markwu: Done!

markwu:
I already squashed all commits to 1.

liuchengxu:
That's not the case. `:Clap grep` is a special one, which is a pure async provider. `:Clap help_tags` is more like `:Clap marks` whose `source` returns a List. When people clear the input, the result list should be the initial complete one. So the current behaviour of `help_tags` is not expected.

markwu:
Then, I have no idea what's going on. The memory cache still there and doesn't changed by anything.

vim-clap

liuchengxu:
I an take a look later, but may not be soon.

markwu:
I just use try and error to find what's going on. It's seems not the bug of the provider itself, it seems the bug, feature or expected behavior from `vim-clap`

```vim
function! myvim#clap#Helptags()
    if empty(s:clap_help_tags_memory)
        if getftime(s:clap_help_tags_cache) > max(map(s:gettagsfiles(), 'getftime(v:val)'))
            let s:clap_help_tags_memory = readfile(s:clap_help_tags_cache)
        else
            let s:clap_help_tags_memory = s:getlist()
            silent! call writefile(s:clap_help_tags_memory, s:clap_help_tags_cache)
        endif
    endif

    return s:clap_help_tags_memory[0:9999]
endfunction
```

I try to play the length of source, if it over 10000, the bug/behavior happened.

I found  you use the number `10000` in some places https://github.com/liuchengxu/vim-clap/search?q=10000&type=Code.

I try to set set

```vim
let g:clap_builtin_fuzzy_filter_threshold = 100000
```

It's useless.

markwu:
I think it's the expected behavior from `let s:async_threshold = 10000` come from the script https://github.com/liuchengxu/vim-clap/blob/master/autoload/clap/impl.vim

liuchengxu:
Good catch. I believe it should be firstly resolved in clap before we continue this provider.

autoload/clap/provider/help_tags.vim
```diff
@@ -0,0 +1,67 @@
...
+  return s:help_tags_memcache
+endfunction
+
+function! s:get_tags_list() abort
+  let tagsfiles = s:get_tags_files()
+
+  let input_dict = {}
+  for tagsfile in tagsfiles
+    for line in readfile(tagsfile)
+      let tag = split(line, "\t")
```
liuchengxu: `tags` are formated by tab, so this should be enough, no need to `matchstr()`.
```vim
      let items = split(line, "\t")
      let tag_subject = items[0]
      if !has_key(input_dict, tag_subject)
        let input_dict[tag_subject] = printf("%-60s\t%s", tag_subject, items[1])
      endif
```

autoload/clap/provider/help_tags.vim
```diff
@@ -0,0 +1,67 @@
...
+      let tag_subject = matchstr(line, '^[^\t]\+')
+      if !has_key(input_dict, tag_subject)
+        let input_dict[tag_subject] = printf("%-60s\t%s", tag[0], tag[1])
+      endif
+    endfor
+  endfor
+  call filter(input_dict, 'v:key !~? "^!_TAG_"')
```
liuchengxu: I'm not sure if this `filter` is neccessary.

markwu: I am not so sure. For ctags, it will generate meta data start with `!_TAG_XXX`. If vim `helptags` won't generate meta data, I think we can remove it.

** meta data example

```
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_OUTPUT_FILESEP	slash	/slash or backslash/
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
!_TAG_PROGRAM_VERSION	0.0.0	/fefbad26/

```

autoload/clap/provider/help_tags.vim
markwu: It's okay for me, I just modified it.

autoload/clap/provider/help_tags.vim
liuchengxu: This provider is for help tags exclusively, so we don't have to care about ctags stuff.

markwu: Done, removed.

liuchengxu:
Please merge the latest clap in your branch. The problem I found before should be fixed now.

Now the thing is to find a more _standard_ cache path for vim-clap. As mentioned above, `stdpath('cache')` is nice for neovim.

The future improvemens could be writing a subcommand in maple for generating the help tags content, so that it willl never be slow, even though this provider is only sluggish on your first run.

markwu:
Okay, I just add a cache layer for Clap.

If user sets `g:clap_cache_dir`, it will use this as default cache folder, if user doesn't set this variable, it will use the following sequence to find the default cache folder:

1. if `has('nvim')`, it will use `stdpath('cache')` + `\clap` as default cache folder
2. if `$XDG_HOME_CACHE` exist, it will use `$XDG_HOME_CACHE` + `\clap` as default cache folder
3. If not 1 and 2, it will use `$HOME`+ `\clap` as default cache folder

So, the developer can use `clap#cache#read(cache_name)` to read cache from file or `clap#cache#write(cache_content, cache_name)`  to cache file.

The cache file will handle `cache_folder` and `cache_file_name` for developer.

autoload/clap/cache.vim
```diff
@@ -0,0 +1,49 @@
...
+let s:path_separator = has('win32') ? '\' : '/'
+let s:clap_cache_dir = get(g:, 'clap_cache_dir', '')
+
+function! clap#cache#folder() abort
+  if empty(s:clap_cache_dir)
+    let user_home = $HOME
+    let user_cache = has('nvim') ? stdpath('cache') :
+          \ exists('$XDG_CACHE_HOME') ? $XDG_CACHE_HOME : 
+          \ user_home . s:path_separator . '.cache'
+    let s:clap_cache_dir = user_cache . s:path_separator . 'clap'
```
liuchengxu: Don't use multiple `expr1` in one single inline, it's not good for the readability.

autoload/clap/cache.vim
```diff
@@ -0,0 +1,49 @@
...
+
+  return cache_file
+endfunction
+
+function! clap#cache#read(cache_name) abort
+  let cache_file = clap#cache#file(a:cache_name)
+  let cache_content = []
+
+  if filereadable(cache_file)
+    let cache_content = readfile(cache_file)
```
liuchengxu: Could return early.

```suggestion
    return readfile(cache_file)
    
    return []
```

autoload/clap/cache.vim
markwu: Done!

autoload/clap/cache.vim
markwu: Done!

autoload/clap/cache.vim
```diff
@@ -0,0 +1,51 @@
...
+
+function! clap#cache#folder() abort
+  if empty(s:clap_cache_dir)
+    let user_home = $HOME
+    if has('nvim')
+      let user_cache = stdpath('cache')
+    elseif exists('$XDG_CACHE_HOME')
+      let user_cache = $XDG_CACHE_HOME
+    else
+      let user_cache = user_home . s:path_separator . '.cache'
```
liuchengxu:
```suggestion
      let user_cache = $HOME . s:path_separator . '.cache'
```

autoload/clap/cache.vim
```diff
@@ -0,0 +1,51 @@
...
+  if filereadable(cache_file)
+    let cache_content = readfile(cache_file)
+  endif
+
+  return cache_content
+endfunction
+
+function! clap#cache#write(cache_content, cache_name) abort
+  let cache_file = clap#cache#file(a:cache_name)
+  call writefile(a:cache_content, cache_file)
```
liuchengxu:
```suggestion
  if filewritable(cache_file)
    call writefile(a:cache_content, cache_file)
  else
    call clap#helper#echo_error('Fail to write cache: '.cache_file)
  endif
```

autoload/clap/cache.vim
```diff
@@ -0,0 +1,51 @@
...
+  endif
+
+  if !isdirectory(s:clap_cache_dir)
+    call mkdir(s:clap_cache_dir, 'p')
+  endif
+
+  return s:clap_cache_dir
+endf
+
+function! clap#cache#file(cache_name) abort
```
liuchengxu: I'm thinking about whether we should provide such an API `clap#cache#location_for(provider_id, fname)` for each provider to avoid the possible naming conflict. We use `top_cache_dir\provider_id\fname` for each cached file path.

autoload/clap/cache.vim
```diff
@@ -0,0 +1,55 @@
...
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+let s:path_separator = has('win32') ? '\' : '/'
+let s:clap_cache_dir = get(g:, 'clap_cache_dir', '')
+
+function! clap#cache#folder() abort
+  if empty(s:clap_cache_dir)
+    let user_home = $HOME
```
liuchengxu:
```suggestion
```

autoload/clap/cache.vim
markwu: We can just extend `clap#cache#file(cache_name)` to `clap#cache#file(provider_d, cache_name)`, It's okay for me. But, how do I know the provider_id is a valid or not?

markwu: Or, don't care about it, just do what user typed?

liuchengxu: The user of this API always knows the provider id, it's generally ensured by the whole system(`autoload` and identifier rule). And `clap#cache#file` is not a general name, for people could merely want to a temp directory instead of a file.

autoload/clap/cache.vim
```diff
@@ -0,0 +1,54 @@
+" Author: Mark Wu <markplace@gmail.com>
+" Description: Cache API for clap.
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+let s:path_separator = has('win32') ? '\' : '/'
+let s:clap_cache_dir = get(g:, 'clap_cache_dir', '')
+
+function! clap#cache#folder() abort
```
liuchengxu: I personally would rename it to `clap#cache#directory` to be more consistent with `g:clap_global_directory`.

autoload/clap/cache.vim
```diff
@@ -0,0 +1,54 @@
...
+  endif
+
+  return s:clap_cache_dir
+endf
+
+function! clap#cache#file(cache_name) abort
+  return clap#cache#folder() . s:path_separator . a:cache_name
+endfunction
+
+function! clap#cache#read(cache_name) abort
```
liuchengxu: I'm unsure if this is neccessary to provide `clap#cache#read` and `clap#cache#write` as we already provided the cache file path via `clap#cache#file()`.

markwu: It's okay for me, readfile() and writefile() is easy enough. The developer only need to compose their cache filename. If you prefer to remove it, I can remove it.

autoload/clap/cache.vim
liuchengxu: Let's just keep these two public APIs then.

markwu: Done!

autoload/clap/cache.vim
markwu: Done!

autoload/clap/cache.vim
markwu: Done!

autoload/clap/provider/help_tags.vim
```diff
@@ -0,0 +1,67 @@
...
+function! s:help_tags_source() abort
+  let help_tags_cache_file = clap#cache#location_for('help_tags', 'help_tags.txt')
+  if empty(s:help_tags_memory_cache)
+    if getftime(help_tags_cache_file) > max(map(s:get_tags_files(), 'getftime(v:val)'))
+      if filereadable(help_tags_cache_file)
+        let s:help_tags_memory_cache = readfile(help_tags_cache_file)
+      endif
+    else
+      let s:help_tags_memory_cache = s:get_tags_list()
+        silent! call writefile(s:help_tags_memory_cache, help_tags_cache_file)
```
liuchengxu:
```suggestion
      silent! call writefile(s:help_tags_memory_cache, help_tags_cache_file)
```

autoload/clap/cache.vim
```diff
@@ -0,0 +1,44 @@
+" Author: Mark Wu <markplace@gmail.com>
+" Description: Cache API for clap.
+
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+let s:path_separator = has('win32') ? '\' : '/'
+let s:clap_cache_directory = get(g:, 'clap_cache_directory', '')
```
liuchengxu: Please also update the doc for `g:clap_cache_directory`.

markwu: What's the order of help files, It seems not follow alphabet order?

liuchengxu: Yes, I just insert the new items randomly, putting them in the right section is enough, but feel free to make it more organized, that could be another PR if you are interested.

autoload/clap/provider/help_tags.vim
markwu: Done!

autoload/clap/cache.vim
markwu: Done!

autoload/clap/cache.vim
```diff
@@ -0,0 +1,44 @@
...
+  if !isdirectory(s:clap_cache_directory)
+    call mkdir(s:clap_cache_directory, 'p')
+  endif
+
+  return s:clap_cache_directory
+endf
+
+function! clap#cache#location_for(provider_id, fname) abort
+  if empty(a:provider_id)
+    let provider_cache_directory = clap#cache#directory()
```
liuchengxu: I suggest not allowing passing a empty `provider_id` in which case it makes no differences to `clap#cache#directory()` and it's not recommended.

markwu: done!

liuchengxu:
Merged, thanks!

PR #258, opened 2020-01-22
Add files multi-select support

liuchengxu:


liuchengxu:
It's trivial to support `git_files`. How is the position in the result list useful?

languitar:
> How is the position in the result list useful?
Apart from the file name that is the only relevant text I can think of to display, because it highlights how well the term matched the file. But maybe the loclist entry order already defines this?

liuchengxu:
Since there is no better optionn, I displayed the basic file properties(modified time and permission) for now.

benwoodward:
This is really good solution, great work! Is there any way to get the results list to display in a horizontal split? I use Nerdtree a lot and the results list divides Nerdtree in half and displays bottom right which means I need to `ctrl-w l, ctrl-w j` to get to it.

liuchengxu:
@benwoodward Have you checked `:h g:clap_open_action`?

benwoodward:
`g:clap_open_action` only seems to affect single files, not multi-selected files. 

Realised that it does what I want it to so long as Nerdtree is closed.

 ✔ 
image

✘ 
image

PR #265, opened 2020-01-26
Introduce g:ClapPrompt for more control of prompt

### LINKED ISSUE #134: Dynamically set the prompt
I want to dynamically set the prompt for each search. Im replicating fzf with source:`fd --type d` sink:`cd`. Then the current dir is shown as prompt in a condensed version.  

Could be a optional provider function `prompt`. I've looked at `g:clap_prompt_format` but that is only sourced once it seems, If that is changed to be checked every time it could be done in `on_enter` I guess?

liuchengxu:
Close #134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
 
 - Add `maple version` to get the detailed maple info and include it in `:Clap debug`.([#262](https://github.com/liuchengxu/vim-clap/pull/262))
 - Add `g:clap_forerunner_status_sign` and deprecate `g:clap_forerunner_status_sign_done` and `g:clap_forerunner_status_sign_running`.
+- Add `g:ClapPrompt` which is Funcref to give more control of the prompt of clap, please see https://github.com/liuchengxu/vim-clap/issues/134#issuecomment-578503522 for the usage.([#265](https://github.com/liuchengxu/vim-clap/pull/265))
 
 ### Changed
 
--- a/autoload/clap/floating_win.vim
+++ b/autoload/clap/floating_win.vim
@@ -147,7 +147,9 @@ function! g:clap#floating_win#spinner.shrink() abort
     let opts = nvim_win_get_config(s:spinner_winid)
     let opts.col += opts.width
     let opts.width = s:display_opts.width - opts.width - s:symbol_width * 2
+    let g:clap#floating_win#input.width = opts.width
     call nvim_win_set_config(s:input_winid, opts)
+    call clap#indicator#repadding()
   endif
 endfunction
 
--- a/autoload/clap/floating_win.vim
+++ b/autoload/clap/floating_win.vim
@@ -137,6 +137,20 @@ function! g:clap#floating_win#spinner.open() abort
   let g:clap.spinner.winid = s:spinner_winid
 endfunction
 
+function! g:clap#floating_win#spinner.shrink() abort
+  let width = clap#spinner#width()
+  let opts = nvim_win_get_config(s:spinner_winid)
+  if opts.width != width
+    let opts.width = width
+    call nvim_win_set_config(s:spinner_winid, opts)
+
+    let opts = nvim_win_get_config(s:spinner_winid)
+    let opts.col += opts.width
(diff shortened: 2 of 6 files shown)


meck:
Great work! I can set the prompt, one issue I tried saving any existing prompt using `on_enter()` and `on_exit()` ie:
``` Vimscript 
function! s:cd.on_enter() abort
  if exists("g:ClapPrompt")
    let s:oldPrompt = g:ClapPrompt
  endif
  let g:ClapPrompt = function('ClapFormat')
endfunction

function! s:cd.on_exit() abort
  if exists("s:oldPrompt")
    let g:ClapPrompt = s:oldPrompt
    unlet s:oldPrompt
  else
    unlet g:ClapPrompt
  endif
endfunction
```

and it seems the prompt is drawn before the call to `on_enter()` I don't know if this is the best way to accomplish this?

liuchengxu:
I think you get confused here. `g:ClapPromt` is a function to generate the prompt when it needs to. You don't have to unlet that. 

```vim
function! s:cd.on_enter() abort
    let s:oldPrompt = clap#spinner#get()
    " Not sure you need to do thing and then call clap#spinner#set(text) here
endfunction

function! s:cd.on_exit() abort
  call clap#spinner#set()
endfunction
```

meck:
Ok, after looking at `spinner.vim` I see, What I am looking to do is to replace the prompt in **one** provider with the `cwd` while keeping the prompt for other providers untouched. 

If I do:
```vim
function! s:cd.on_enter() abort
  call clap#spinner#set('foo')
end-function
```
The prompt resets when typing and also I can't use `%spinner%`.

I wrongly assumed `g:ClapPrompt` was checked for every time the prompt was drawn, but I see its only checked on initial load.

What I was looking for is a per provider `ClapPrompt`, not a global one.

liuchengxu:
>I wrongly assumed g:ClapPrompt was checked for every time the prompt was drawn, but I see its only checked on initial load.

It's indeed initialized only once, but will be evaluated every time the prompt needs to be drawn.

As I have mentioned, being a function `g:ClapPrompt` is able to do anything you want to:

```vim
function! ClapFormat() abort
  " Only use your own format for files provider,
  " otherwise use the default prompt.
  if g:clap.provider.id ==# 'files'
    if exists('g:__clap_provider_cwd')
      let cwd = g:__clap_provider_cwd
    else
      let cwd = getcwd()
    endif
    return '%spinner% '.pathshorten(cwd).':'
  else
    return ' %spinner%%forerunner_status%%provider_id%:'
  endif
endfunction

let g:ClapPrompt = function('ClapFormat')
```

meck:
Thanks for all your help and assistantance! The problem is that I seek to do is, set the prompt from within a provider, without changing anything defined in `.vimrc`. What I tried doing now is something like this:
``` vim
if exists("g:ClapPrompt")
  let s:prevClapPrompt = funcref("g:ClapPrompt")
endif

function! ClapFormat() abort
  if g:clap.provider.id ==# 'cd'
    if exists('g:__clap_provider_cwd')
      let cwd = g:__clap_provider_cwd
    else
      let cwd = getcwd()
    endif
    return '%spinner% '. pathshorten(fnamemodify(expand(cwd), ':~:s?/$??')) . ' '
  elseif exists("s:prevClapPromt")
    call s:prevClapPrompt()
  else
    if exists("g:clap_prompt_format")
      return g:clap.prompt_format
    else
      return ' %spinner%%forerunner_status%%provider_id%:'
    endif
  endif
endfunction

let g:ClapPrompt = function('ClapFormat')
``` 
However that only works if the provider is the first one opened after launch, so `s:PromptFn()` is defined with that one

liuchengxu:
`g:ClapFormat` is a user-level setting, it should not be used like that.

If you want to change the prompt in the provider temporarily, just use `clap#spinner#set(text)` then, the prompt will be reset to the value of `s:PromptFn()` later, or you can use `clap#spinner#refresh()` to reset it to `s:PromptFn()` manually.

liuchengxu:
@meck  I add  a provider level `prompt_format` in #319 and use it for showing the current executive of the tags returned by vista.vim which is changed dynamically. You might want to take a look.

PR #275, opened 2020-01-29
Add readme section on clap_layout

zakaprov:
I think it's worth including this variable in the readme. So far, it was only documented in the changelog and it had a major impact on the UX (changing vim-clap position to be relative to the active window by default).

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,12 @@ Command                                | List                                   
 
 - `g:clap_open_action`: Dict, `{ 'ctrl-t': 'tab split', 'ctrl-x': 'split', 'ctrl-v': 'vsplit' }`, extra key bindings for opening the selected file in a different way. NOTE: do not define a key binding which is conflicted with the other default bindings of vim-clap, and only `ctrl-*` is supported for now.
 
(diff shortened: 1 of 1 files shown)


liuchengxu:
Actually I have documented it in the vim help doc, see `:h g:clap_layout`.

zakaprov:
Yes, that's where I eventually learnt about it, though it took me some time (I started looking in the GitHub readme). I just thought it could be a shortcut for some people, since the section on global variables is already there.

liuchengxu:
I have not added the new items in global variables section of GitHub README for some time, for I don't want to maintain the README and help doc at the same time. But since you have made this PR, I can just accept it :(.

zakaprov:
I see, fair enough. In that case I think it could be helpful to mention this at the beginning of the global variables section (e.g. saying the list might not be up to date and redirecting to the help docs). I can open another PR for that if you want. :D

liuchengxu:
>See :help clap-options for more information.

I mention this at the ending of each section actually :(.

zakaprov:
True, what I meant was to underline that the list is not complete and may be outdated. This might be an isolated issue, but I missed that mention at the end of the section. :x

PR #290, opened 2020-02-02
Add g:clap_project_root_markers

### LINKED ISSUE #283: Project root for mono-repos
First, thanks for vim-clap ! Very promising project to improve over `ctrlp` :)

It will be nice to be able to restrict the how far up the directory tree the search goes. We use a monorepo and when we work in one project, we don't want to get results from other project in the git repo.

I think the way ctrlp handle this works pretty well. Having a `.vimclapstopper` file that indicates were to stop searching up the directory tree. 

**Describe alternatives you've considered**
Since the solution mentioned above seems to work well, I haven't thought much of other solution. That is the cleanest one that comes to my mind.

I'm currently the `let g:clap_disable_run_rooter = v:true` option and it's a fine workaround but it'd be great not having to keep yet another thing in mind :P

liuchengxu:
Close #283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ CHANGELOG
 
 ### Added
 
+- Add new provider `:Clap search_history` by @markwu. ([#289](https://github.com/liuchengxu/vim-clap/pull/289))
+- Add new provider `:Clap maps` by @markwu. ([#293](https://github.com/liuchengxu/vim-clap/pull/293))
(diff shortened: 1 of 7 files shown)


markwu:
I saw the document, the default value is `['.root', '.git', '.git/']`. I am a little bit confused, what's the different between `.git` and  `.git/`.

If I want to add `.svn` into my root marker support, should I also add `.svn/`?

liuchengxu:
`.git/` is for gitsubmodule.

markwu:
Ah, I see. Thanks!

PR #308, opened 2020-02-05
doc: Update the "Download the pre-built binaries"

yaegassy:
Since `clap#helper#download_binary()` has been changed to `clap#installer#download_binary()`, "Install.md" also needs to be updated.


liuchengxu:
Thanks!

Issue #129, opened 2019-11-19
Allow using non-english text


habamax:
**Is your feature request related to a problem? Please describe.**
Impossible to grep for a russian text and probably open files with russian file names.

**Describe the solution you'd like**
Allow entering text either using keymap (C-^) or non-english text with OS language switch

liuchengxu:
Vim's popup can't get the focus like NeoVim's floating_win, so the search text window and cursor of popup are mocked by vim-clap, which has to remember the emulated cursor position manually. 

https://github.com/liuchengxu/vim-clap/blob/3e16a996027ba7e7924b98eff34df9b634409c28/autoload/clap/popup.vim#L294-L307

vim-clap currently only responses to the ASCII printable characters:

https://github.com/liuchengxu/vim-clap/blob/3e16a996027ba7e7924b98eff34df9b634409c28/autoload/clap/popup.vim#L499-L502

Supporting non-english text means to support multi-byte text, which involves the `s:cursor_idx` calculation mainly. With the right cursor index, the multi-byte text display should be fine, but you have to also take care of these motions, i.e., `<C-A>`, `<C-B>`, etc. 

https://github.com/liuchengxu/vim-clap/blob/3e16a996027ba7e7924b98eff34df9b634409c28/autoload/clap/popup.vim#L458-L470

It's not a job I like, and it's critical for me, so I may won't spend time on this issue. PR is welcome.

PR #320, opened 2020-02-12
Support multibyte input for popup filter

Bakudankun:
Currently clap does not accept non-ascii characters for filtering on Vim (#129). I enabled that by using `strchar*()` functions instead of byte index. `s:cursor_idx` has been changed to indicate character index instead of byte index.
Since `strcharpart()` does not throw error for out-of-bound indexes, the source have gotten simpler.

Also, I fixed that the cursor in popup was not highlighted properly if the cursor contains multi-byte characters.

### DIFF
--- a/autoload/clap/popup/move_manager.vim
+++ b/autoload/clap/popup/move_manager.vim
@@ -150,7 +150,7 @@ call s:define_open_action_filter()
 
 function! s:move_manager.printable(key) abort
   let s:input = s:strpart_input(0, s:cursor_idx).a:key.s:strpart_input(s:cursor_idx)
-  let s:cursor_idx += strchars(a:key, 1)
+  let s:cursor_idx = strchars(s:strpart_input(0, s:cursor_idx) . a:key, 1)
 
   " Always hold a delay before reacting actually.
   "
--- a/autoload/clap/popup/move_manager.vim
+++ b/autoload/clap/popup/move_manager.vim
@@ -27,7 +27,7 @@ endfunction
 
 function! s:move_manager.ctrl_f(_winid) abort
   let s:cursor_idx += 1
-  let input_len = strlen(s:input)
+  let input_len = strchars(s:input, 1)
   if s:cursor_idx > input_len
     let s:cursor_idx = input_len
   endif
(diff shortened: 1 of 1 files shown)


liuchengxu:
Awesome! I just tested and it works well. @habamax  You might also want to try this patch.

habamax:
Thx! I will check it out!

habamax:
I get an error:

image

liuchengxu:
@habamax How can we reproduce it?

Bakudankun:
I noticed that some multi-byte input can pass invalid character to the filter.
`｝` does not, but `｛` passes extra bytes after the character.
`、` and `。` get extra bytes in mid of their byte sequences.

Maybe a bug of vim?

liuchengxu:
I also see the similar errors on my side. Not sure if it's bug of vim, handling the raw key manually is just not a fun job :(.

Bakudankun:
Another possible method is to use command-line.
Though I don't know much, with combination of `input()`, `getcmdline()` and `+timers` (or maybe `CmdlineChanged` autocmd), you can get what user input into command-line asynchronously. `cmap` can be used to set mappings.
I've heard that [lambdalisue/filta.vim](https://github.com/lambdalisue/filta.vim) uses this method. [Here](https://github.com/lambdalisue/filta.vim/blob/master/autoload/filta/filter.vim#L10) seems to be its starting.

It may also be a hard work, and command-line cannot be floating, but at least no need to mock input window.

liuchengxu:
Using the command-line way is too tricky and a mess. The perfect solution is the native focus support of vim's popup, but I'm unsure if it'll happen eventually.

habamax:
> Using the command-line way is too tricky and a mess. The perfect solution is the native focus support of vim's popup, but I'm unsure if it'll happen eventually.

Leaderf somehow manages non-ascii in popups, not sure how hacky it is though

habamax:
> 
> 
> @habamax How can we reproduce it?

I have just opened `:Clap files` change windows input to russian and start typing.

liuchengxu:
Anyway I have sent the feature request in https://github.com/vim/vim/issues/5639, I think it's better be resolved by vim itself.

Bakudankun:
I confirmed that the issue I noticed has been fixed at the very latest vim v8.2.0329 (vim/vim@ec084d3356cc2a23f2f44707810c38ab590d2e66).
And, I fixed a bug that composing characters breaks the move manager.

liuchengxu:
@Bakudankun  Nice work!

@habamax You can can try branch again.

habamax:
@liuchengxu works like a charm! thank you!

liuchengxu:
@Bakudankun is that hero, @habamax . 

Merged, thanks @Bakudankun !

PR #330, opened 2020-02-24
Ensure the long matched elements always at least partially visible

### LINKED ISSUE #317: Long matched results can cause the matched portion to not be visible

When matching entries that are really long, it is possible for the matched/highlighted portion to exist outside of the visible window. e.g., when using `:Clap files` to search by filename in a project that has a deeply-nested folder structure:

image

This can make it difficult to understand which entry is actually being selected.

It would be nice if vim-clap could truncate or justify the entry so that the highlighted portion is visible within the window:

image

Thank you! vim-clap has been really nice to use :)

liuchengxu:
The future work is to optimize for maple grep as well.

Close #317

截屏2020-02-24下午5 14 35


YellowKirby:
This is working pretty well for me, thank you! One thing that seems a little weird is that the input line is also scrolling, even though it seems like there is room to show more of the input. 

image


In this screenshot: terminal is 45 columns wide, I have not made any configuration changes to how vim-clap opens the popup. I've typed `filea`, but only the letter `a` is visible.

liuchengxu:
I can only ensure it won't be invisible tototally :(, no guarantee about occupying the whole space, feel free to improve in https://github.com/liuchengxu/vim-clap/pull/330/files#diff-bc69e7d07f276d599a29af5c04661940R111 .

Issue #337, opened 2020-02-25
Icons in :Clap files like in :Clap grep


GoldsteinE:
`:Clap grep` shows filetype devicons and `:Clap files` doesn't.

**Describe the solution you'd like**
`:Clap files` (and other providers, like `:Clap gfiles`) should also have option to show devicons.

**Describe alternatives you've considered**
Maybe icons can be implemented not on provider level, but in Clap itself.

liuchengxu:
`files` like icon support is only avaiable when you are using the Rust extension, see #191. I don't have to time to optimize for the pure-vimscript implementation as that affects the performance. Checkout `convert` property of provider and the example in `:Clap grep` if you are interested. The icon support has be done in each provider, clap, as the framework itself, has no idea to implement a generic icon support.

PR #340, opened 2020-03-01
Block cursor for popup filter

Bakudankun:
* Make the cursor in popup filter block-shaped when `g:clap_popup_cursor_shape == ''`.
* I feel it deserves to be default, so make it default.
* I noticed `<Del>` does not work in popup filter, so make it work as it works on neovim.

before:
before

after:
after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ CHANGELOG
 - Now you can use `:Clap grep ++query=@visual` to search the visual selection. ([#336](https://github.com/liuchengxu/vim-clap/pull/336))
 - Ensure the long matched elements from the filter always at least partially visible. ([#330](https://github.com/liuchengxu/vim-clap/pull/330))
 - Use file name as the preview header for `Clap grep`, `Clap marks` and `Clap jumps`.
+- Use block-style cursor for vim's popup when `g:clap_popup_cursor_shape == ''`. The default value has been changed to that.
 
 ## [0.8] 2020-02-21
 
--- a/autoload/clap/popup/move_manager.vim
+++ b/autoload/clap/popup/move_manager.vim
@@ -195,7 +195,8 @@ function! s:hl_cursor() abort
 endfunction
 
 function! s:mock_input() abort
-  let input = s:strpart_input(0, s:cursor_idx).s:cursor_shape.s:strpart_input(s:cursor_idx)
+  " The trailing space is for block cursor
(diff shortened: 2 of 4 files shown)


liuchengxu:
This looks nice!

P.S. I notice a glitch in your screencast. The highlight of fuzzy matched items, e.g., `clap`, are shifted two chars to the right. I don't see this issue locally. Are you using the latest version?

Bakudankun:
> P.S. I notice a glitch in your screencast. The highlight of fuzzy matched items, e.g., `clap`, are shifted two chars to the right. I don't see this issue locally. Are you using the latest version?

Yes. This occurs from before my first PR. I'm using Windows Vim, so that may be a cause.

Bakudankun:
This glitch of highlights also seems to occur in neovim if candidates include multi-byte characters. It looks similar to that the cursor in popup filter had not been highlighted properly if the cursor includes multi-byte characters.

liuchengxu:
So it's a GUI related issue? I see no problem with kitty, iterm2 and alacritty using both neovim and vim on macOS.

截屏2020-03-01下午3 43 52

I firstly encounter this problem when I was introducing the icon support and adjusted them manually. Try to adjust the offset https://github.com/liuchengxu/vim-clap/blob/master/autoload/clap/highlight.vim#L1-L38. I never test windows, so I don't know if these tweaks also work in your case.

Bakudankun:
Well, I'm still not sure why this happen, but I gathered some clues.

* With maple, the glitch in the above gifs does not occur. Glitches still occur for multi-byte candidates.
* The glitch seems to occur only on "files" and "gfiles". Note that they does not show icons if maple is not available.
* When I make `s:builtin_fuzzy_idx_offset()` always returns 0, the glitch was dismissed on "files" and "gfiles". However on other providers highlights shift in the opposite direction.

liuchengxu:
I believe the root cause is the current fuzzy filter(whether the Python or Rust implementation) can't handle the multi-byte chars appropariately.

Icon support is not added to the vimscript implementation, see https://github.com/liuchengxu/vim-clap/issues/337#issuecomment-590853187.

Since that glitch is not clear yet. We could merge this first and resolve it in future. Thanks for this PR, the block-style cursor looks great!

liuchengxu:
@Bakudankun  I rewritethe icon highlight offset in #389, you can try it if you still see the misordered fuzzy matched highlights.

Issue #302, opened 2020-02-05
Support dynamic update in maple for grep/files provider


liuchengxu:
Currently maple only responses when the job is finished, i.e., there is no dynamic update for the time-consuming jobs. We have a spinner in the vim window, that's good, but not perfect. We have introduced a new rpc-based provider in #272, that means all the work required on the vim side has been done, now the job is to implement a dynamic update for running a system command in maple, but it's obviously not easy. It sounds like the core logic of skim without drawing the UI.

1. Start the maple service and run the command on start.
2. User type inside vim.
3. Send the input event to maple via job channel.
4. Response the input event and send back the update periodly.

-------------

- [x] files provider will support dyn filter after #364 is merged.
    - And any provider can benefit from it if the `source` is a String or a function that returns a String.
    - From what I have seen, neovim works great as always with the dyn filter, but vim still behaves bad due to the linewise job invocation. I believe using a timer to reading the output periodly can help, but I'm a neovim user, this timer improvement will be indefinite.
- [ ] new grep2 provider #383, with cache and dynamic refresh support.

ImmemorConsultrixContrarie:
Just to make sure I know what I'm talking about, before I actually start talking about it: what is "maple"? A Rust or a Python extension?

liuchengxu:
@ImmemorConsultrixContrarie  https://github.com/liuchengxu/vim-clap/blob/master/Cargo.toml#L2 The silly Rust external binary for vim-clap :(.

ImmemorConsultrixContrarie:
One more question: there's a big uncommented matcher in `Maple::run()`, which part of this matcher should provide dynamic results?
@liuchengxu

liuchengxu:
The dynamic update is still unimplemented, but I think the current filter/matcher can be reused.

ImmemorConsultrixContrarie:
> 
> 
> The dynamic update is still unimplemented, but I think the current filter/matcher can be reused.

No-no-no, There's a match for commands, and I don't understand which command you want to change.
```
        match self.command {
            Cmd::Version => {
                version();
            }
            Cmd::RPC => {
                crate::cmd::rpc::run_forever(std::io::BufReader::new(std::io::stdin()));
            }
            Cmd::Filter { query, input, algo } => {
                let source = input.map(Into::into).unwrap_or(Source::Stdin);
                crate::cmd::filter::run(
                    query,
                    source,
                    algo,
                    self.number,
                    self.enable_icon,
[...]

Issue #292, opened 2020-02-02
RAM overload with more than 600K results


SqrtMinusOne:
**Environment (please complete the following information):**
 - OS: Linux Mint 19.2
 - NeoVim version: 0.4.3
 - vim-clap version: commit f459551
 - Have you reproduced with a minimal vimrc: No
 - Have you updated to the latest plugin version: Yes
 - Have you upgraded to/compiled the latest Rust binary: Yes

**Describe the bug**
I frequently use the global fuzzy-finder. Sometimes ripgrep outputs an unholy amount of results, e.g. in one of the projects I'm currently working on an output of
`rg -H --no-heading --vimgrep --smart-case "$cookie" | wc -l`
would be `619660` (619660 lines in the result). I wanted to find places where I use `$cookie` and forgot that `$` is a special symbol in regex, so certain lines were repeated multiple times.

Ripgrep can handle this, but vim-clap consumes a lot of RAM while receiving such a large input. 200K lines is fine, but 600K lines is more than my computer can handle.

In essence, if I type something wrong in `Clap grep`, I risk crashing my PC.

**Clap debug**
```
has cargo:             1
has maple:             /home/pavel/.local/share/nvim/plugged/vim-clap/target/release/maple
maple info:            version 0.1.7, built for x86_64-unknown-linux-gnu by rustc 1.40.0 (73528e339 2019-12-16).
has +python3:          1
has py dynamic module: 0
Current FileType:      startify
Third Party Providers: []

Global Options:
let g:clap#autoload_dir              = '/home/pavel/.local/share/nvim/plugged/vim-clap/autoload'
let g:clap#provider_alias            = {'gfiles': 'git_files', 'hist:': 'command_history'}
let g:clap_disable_bottom_top        = 0
let g:clap_enable_icon               = 1
let g:clap_forerunner_status_sign    = {'done': '•', 'running': '!'}
let g:clap_layout                    = {'relative': 'editor'}
let g:clap_no_matches_msg            = 'NO MATCHES FOUND'
let g:clap_open_action               = {'ctrl-v': 'vsplit', 'ctrl-x': 'split', 'ctrl-t': 'tab split'}
let g:clap_search_box_border_style   = 'nul'
let g:clap_search_box_border_symbols = {'nil': ['', ''], 'curve': ['', ''], 'arrow': ['', '']}

Provider Variables:
[]
```
**To Reproduce**
[...]

PR #353, opened 2020-03-20
An example of dynamic results provider, using stdout

ImmemorConsultrixContrarie:
Something like this could be used for #302. I can't test the code, because I'm too lazy to run it without vim, and I can't run it with vim. But I tried to keep code simple enough.

I don't think that this PR could be simply merged, because I just pulled it together to the point where `cargo check` stopped producing errors. And thus I'm making this PR a draft.


crates/fuzzy_filter/src/lib.rs
```diff
@@ -83,12 +84,646 @@ pub fn fuzzy_filter_and_rank(
...
+    //XXX  }
+    //XXX  ```
+    //XXX
+    //XXX This really looks like a job for some simple script (python, bash),
+    //XXX that will generate a .rs file with the proper Rust code.
+    //XXX Even though this could be easily done manually once, updating such tables
+    //XXX and matches is very prone to errors and it's not something
+    //XXX you want to do, if you can just add one line to script instead.
+    //XXX But if don't plan to update those things at all, that could be done without script once.
+    const EXACTMATCH_MAP: [(&'static str, &'static str); 33] = [
```
liuchengxu: I have made #354 to split out `src/icon.rs` as another crate. Can you pick out the icon related code into another PR after #354 is merged? That would make this PR cleaner. The icon optimization looks great to me.

ImmemorConsultrixContrarie:
I still need to know what you want to do with RPC on higher level, so can't connect that function with your rpc module, and still can't test it, so I don't know if it works at all.

liuchengxu:
@ImmemorConsultrixContrarie Can we make another PR for these improvements that not related the dynamic update part, that would make the review easier.

Here is the dynamic update in my mind: given the query and list of candidates, do the filtering and ranking, this could take a while, so we need to update the top N scored candidates and print the top N to stdout periodically, e.g., print new top N every 50ms if they changes. For the test purpose at the moment, I believe you could write a test to mock this process? I can take the rest of work if this core part is done.

ImmemorConsultrixContrarie:
> Can we make another PR for these improvements that not related the dynamic update part, that would make the review easier.

Welp, theoretically, yes. But I'm too lazy to do it, so no. You can `git rebase` it as you like, though.

crates/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[workspace]
+
+members = [
+    "extracted_fzy",
+    "fuzzy_filter",
+    "icon",
+]
```
ImmemorConsultrixContrarie: I'd like to add Maple binary into that workspace too, but this could break some other code, and I wanted to cooperate with someone who knows what code could get broken (or at least test everything and find out if something gonna break).
@liuchengxu

liuchengxu: @ImmemorConsultrixContrarie 

> I'd like to add Maple binary into that workspace too.

I'm ok with that and I can test it.

To avoid the conflicts, I think it's better to start the new refactoring after #356 is merged, I have cherry-picked the code that are unrelated to the dynamic update in this PR into #356.

ImmemorConsultrixContrarie:
Okay, the last commit breaks something, as I can't compile maple now with "error: couldn't read maple\build.rs".
@liuchengxu

liuchengxu:
@ImmemorConsultrixContrarie  `mv build.rs crates/maple` should fix that problem. But I don't do recommend removing all the maple code to `crates/maple` as that also breasks the vimscript code. I prefer to make `crates/maple` a lib, keep the current `src/main.rs` to call the `run()` funnction from `crates/maple`, that will not break any thing.

ImmemorConsultrixContrarie:
> 
> 
> @ImmemorConsultrixContrarie `mv build.rs crates/maple` should fix that problem. But I don't do recommend removing all the maple code to `crates/maple` as that also breasks the vimscript code. I prefer to make `crates/maple` a lib, keep the current `src/main.rs` to call the `run()` funnction from `crates/maple`, that will not break any thing.

So, three files: "vim-clap/build.rs" with "vim-clap/src/main.rs" and "vim-clap/Cargo.toml". Should it look like this?

liuchengxu:
> So, three files: "vim-clap/build.rs" with "vim-clap/src/main.rs" and "vim-clap/Cargo.toml". Should it look like this?

Yes! And we could name the lib `crates/maple` to `maple_cli` or other something. Then nothing will be broken I think.

crates/maple_cli/src/cmd/filter.rs
```diff
@@ -0,0 +1,499 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    // This is a very time-consuming test,
+    // results of which could be proved only be inspecting stdout.
+    // Definetly not something you want to run with `cargo test`.
+    #[ignore]
+    fn dynamic_results() {
```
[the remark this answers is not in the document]
liuchengxu: Have not got time to look at the test as I have to merge the latest mater/rebase this PR as it has too much code that are unrelated. Will let you know when I tried the test.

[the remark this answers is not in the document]
ImmemorConsultrixContrarie: > 
> 
> Have not got time to look at the test as I have to merge the latest mater/rebase this PR as it has too much code that is unrelated. Will let you know when I try the test.

It's just `git clone https://github.com/ImmemorConsultrixContrarie/vim-clap` and `cd vim-clap/crates` with `cargo test -- --ignored --nocapture`.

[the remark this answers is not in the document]
liuchengxu: @ImmemorConsultrixContrarie  Just grab the code and run the test, it just works the way that I have always wanted for dynamic filter, awesome job! :+1: 

I suggest using `Source::File(random_generated_data_file)` for the test purpose as we could compare the final result of dynamic version and the non-dynamic version and that's one of the real use case, and it's more readable personally. But it's ok if you're too lazy for this, I can take the test job anyway. For me, I use this locally:

```bash
$ cd / && fd --type f > ~/root.txt
$ # Use Source::File(`~/root.txt`) in the test
```

Once we get this big feature merged, we can also expand the dynamic feature further. The ultimate imagtination is the `Source` can be generated in async, like we spawn the command `fd --type f` in a thread and use its stdout as the source of our filter, do the filtering in another thread.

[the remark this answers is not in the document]
ImmemorConsultrixContrarie: > I suggest using `Source::File(random_generated_data_file)` for the test purpose as we could compare the final result of dynamic version and the non-dynamic version and that's one of the real use case

Yeah, but this should be really big file and I don't have any. And you can just make this generator determined, if you write `let bytes = 123;` instead of time-dependent bytes.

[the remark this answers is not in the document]
ImmemorConsultrixContrarie: Okay, if it works, I guess this draft should be finally merged with master branch, so next commits won't be that painful to work with, so I'm gonna mark this draft as finally ready for review.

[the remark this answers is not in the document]
liuchengxu: @ImmemorConsultrixContrarie Actually it's easy to get the testdata it I have used it before :(.

```bash
$ vim-clap/test/bench/python
$ bash fetch_testdata.sh
$ # Now you'll see testdata.txt
```

This test file has about 850K lines and has beenn ignored in .gitignore.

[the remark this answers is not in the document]
liuchengxu: @ImmemorConsultrixContrarie Since you have marked this PR reviewable, can you please resolve the conflicts also? Except the icon related code, most of them should just take yours.

ImmemorConsultrixContrarie:
Resolved.
@liuchengxu

liuchengxu:
I have cherry-picked the last several commits unrelated to the dynamic update part into the master, you can merge the latest master again to resolve the confilicts, and then I think we could focus on the core part of this PR. @ImmemorConsultrixContrarie

ImmemorConsultrixContrarie:
cc @liuchengxu

crates/maple_cli/src/cmd/filter.rs
```diff
@@ -70,3 +70,409 @@ pub fn blines(
...
+        if let Ok(()) = if_ok_return {
+            return buffer;
+        };
+
+        // Now we have the full queue and can just pair `.pop_back()` with `.insert()` to keep
+        // the queue with best results the same size.
+        let mut past = std::time::Instant::now();
+        iter.for_each(|(text, score, indices)| {
+                // Best results are stored in front.
+                //XXX I can't say, if bigger score is better or not. Let's assume the bigger the better.
```
liuchengxu: I can assure you that bigger score is better in our secerior.

crates/maple_cli/src/cmd/filter.rs
```diff
@@ -70,3 +70,409 @@ pub fn blines(
         winwidth,
     )
 }
+
+/// Return the ranked results after applying fuzzy filter given the query string and a list of candidates.
+pub fn dyn_fuzzy_filter_and_rank<I: Iterator<Item = String>>(
```
liuchengxu: I think it's worthy of moving all the dyn_* part into a new module under `filter`, putting the const items on the top, moving the trait/macro outside the functions.

crates/maple_cli/src/cmd/filter.rs
```diff
@@ -70,3 +70,409 @@ pub fn blines(
...
+                list.filter_map(|line| {
+                    scorer(&line).map(|(score, indices)| (line, score, indices))
+                }),
+                number,
+            ),
+        };
+
+        let ranked = filtered;
+
+        let payload = ranked.into_iter().take(number);
```
liuchengxu: I see some code repeations here. I can refactor it later if you are focusing the dyn part.

ImmemorConsultrixContrarie: I was just putting all things into one function to rewrite it as dynamic, and this renaming shows a borderline between old functions.

And I don't think I'll be doing any other big changes to this particular function, because now it has no unsafe and works. But, for example, if you want to search in all files in a folder and its subfolders *in parallel*, the filtering and dynamic updates should be done other way.

liuchengxu: >if you want to search in all files in a folder and its subfolders effective and in parallel, the filtering and dynamic updates should be done other way.

clap iterself does not do the job of searching, but relays on the tools like `fd`, `ripgrep`, which already does the best in terms of searching. What clap mainly does is to do the filtering on the data of `fd`/`ripgrep` produces, so we don't have to worried about this point.

ImmemorConsultrixContrarie: > 
> 
> > if you want to search in all files in a folder and its subfolders effective and in parallel, the filtering and dynamic updates should be done other way.
> 
> clap iterself does not do the job of searching, but relys on the tools like `fd`, `ripgrep`, which already does the best in terms of searching. What clap mainly does is to do the filtering on the data of `fd`/`ripgrep` produces, so we don't have to worried about this point.

So, you make a lot of serializing-deserializing between fd / ripgrep and python module, and then again ser-deser from python to maple. Or am I wrong?

liuchengxu: Nope. Actually the Python part is another topic. The Python dynamic module is only used when all the data is ready and for the small dataset as:

1. The Python filtering  is sync which can block vim if there are too much data.
2. There is no need to spawn a `job` in vim to communiate with the external binary maple.

When the dataset is huge, e.g., 1 million lines, I just use the `std::process::Command` to execute the fd/ripgrep command, gather their output and then do the filtering. I know it's not good enought, but at least vim won't be blocked as all the heavy job runs externally.

So my final goal for maple is to run the job like `fd`/`ripgrep` for producing data in one thread, reading the output in another thread and do the filtering, which is mostly what current popular fuzzy finders like `fzf`/`skim` do. The difference between `fzf`/`skim` and vim-clap is that `fzf`/`skim` are essentially TUI programs even they can also used in vim, drawing the UI by theirselves, for vim-clap, it uses the native Vim UI stuffs.

ImmemorConsultrixContrarie: > I just use the `std::process::Command` to execute the fd/ripgrep command, gather their output and then do the filtering

And what is the output? Is it files' handles, lines, or something else?

To not talk about this unfounded, that's how I see effective fuzzy-filtering of many files with dynamic results:

1. There's always a number-cap for results. If it's not provided by the user, it's somethings like 500, 1000 or 2000, some number of results that is hard to read for humans but easy to store for computers (and if that's not enough for the user, he can provide a bigger number and run filtering again). This is almost always better than absence of such cap, and always better for big numbers of total results.
2. Bigger tasks are parallel, lesser aren't. Instead of parallel reads of lines in one file, which is not very effective, threads share the pool of files or lines and take some files or many lines at once, then process those in old one-threaded manner, synchronizing results from time to time (how and when to synchronize is a lower-level details, that aren't clear just now). Now, instead of many short-living threads, that fight for OS resources each time they are created, we have some long-living threads, that are almost always working.

Simply saying, all this optimizing is around less number of reallocations and more effective use of threads.

liuchengxu: > And what is the output? Is it files' handles, lines, or something else?

Just the `Output` of `std::process::Command`, see https://github.com/liuchengxu/vim-clap/blob/master/crates/maple_cli/src/light_command.rs#L78-L90. I'll then convert this `Output` to lines for the filtering.

It's ok to always enforce a number-cap as that's the typical use case, and the screen size is limited, people normally won't check out more than dozens of items.

liuchengxu: @ImmemorConsultrixContrarie Yes, one vim instanace has only one maple-bin running, but the maple process can be created and killed multiple times per vim instance context.

ImmemorConsultrixContrarie: > Just the `Output` of `std::process::Command`, see https://github.com/liuchengxu/vim-clap/blob/master/crates/maple_cli/src/light_command.rs#L78-L90. I'll then convert this `Output` to lines for the filtering.

Yeah, I see how you get `&[u8]` from the output, but can't find where it's passed into fuzzy-filter, only more `println!`s.

liuchengxu: hmm, it's a little complex. Due to the missing of dynamic update, gathering the output of `Command` and filtering given the query is splitted into two steps with some temp naive optimization, such as writting to a temp file. But that would be unnecccary with the dyamic update, now we can just feed `&[u8]` into the filter.

ImmemorConsultrixContrarie: So, do we change the current filterer, so it works with `&[u8]` and always-capped number, and then land this filterer, even though it's one-threaded right now; or do we write a proper multithreaded filterer, that is focused on working with `&[u8]` and always-capped number?

Or even more epic task: merging ripgrep / fd code with fuzzy-filter, so whenever you find a line, you fuzzy-filter it in place and either store it or throw away. This is better for both thread-usage and RAM problems like #292, as there will always be a cap number and no more than that number of lines will be ever stored.

ImmemorConsultrixContrarie: By the way, rg/fd search for regex, but fuzzy-filter isn't regexed, so linefinder for fuzzy-filter could be very simple. Or am I wrong and you can fuzzy-search regexes?

liuchengxu: >merging ripgrep / fd code with fuzzy-filter,

That's crazy 🙉 , that could be hard to maintain I think.

I think it's a good choice to write a multithreaded filter, one thread A for producing data given the fd/ripgrep command, the other for filtering on the data produced by thread A and notifying the result periodly.

liuchengxu: fuzzy-search regexes is unsupported at the moment, but it sounds like a great addition!  @ImmemorConsultrixContrarie

ImmemorConsultrixContrarie: > 
> 
> fuzzy-search regexes is unsupported at the moment, but it sounds like a great addition! @ImmemorConsultrixContrarie

I can't even imagine how this thing could be implemented (apart from very slow implementation, when you make a regex into all strings it could represent and then fuzzy-search for each in each line) . Now THIS is crazy, man.

So, right now you can search for regex OR fuzzy-search some letters in line. And the latter task doesn't need any regex search, actually, just some uses of [memchr](https://docs.rs/memchr/2.3.3/memchr/). And I can merge such a simple linefinder with fuzzy-filterer for sure. And [ignore](https://docs.rs/ignore/0.4.13/ignore/) crate will help with recursive directory search.

So, if you don't plan writing regex fuzzy-finder yourself, I'll just stick to this very simple plan, that will optimize RAM usage, thread usage and probably make fuzzy-search simply faster.

liuchengxu: https://github.com/junegunn/fzf#search-syntax Although fzf does not support full regexp based searhcing, it has a extended search mode, and I think it's useful, I might also want to introduce it in clap in the long term.

ImmemorConsultrixContrarie:
Short thought about `icon` while I'm playing with new file-line filter:

It's probably better to hold a static reference to const array instead of array itself, because all arrays are `Copy` and this could accidentally do some unnecessary copying of big array. See [std unicode src](https://doc.rust-lang.org/src/core/unicode/unicode_data.rs.html#635) for example of how this could be done. Not doing this myself, because I'm terrible with everything except Rust (including Python, lol).

@liuchengxu

liuchengxu:
@ImmemorConsultrixContrarie That's neat, I can do this :(.

liuchengxu:
The workflow of clap is as follows:

```
[vim] User invokes clap

      🡣  {command}: [maple] runs {command},
                    collect its output as DATA of the following filterings.


[vim] User inputs something

      🡣  {query1}: [maple] filters on DATA given {query1} -> notify results


[vim] User inputs something

      🡣  {query2}: [maple] filters on DATA given {query2} -> notify results


[vim] User inputs something

      🡣  {query3}: [maple] filters on DATA given {query3} -> notify results


      Keep changing the query
      ......

[vim] User exits.
```

1. User invokes clap from vim, then I get the source command to run it in maple and collect the output as the dataset we  are about to filter on. I _store_ the output temperiorily so that do not have to run the command again.

2. User keeps changing the query string, maple do the filtering given the query and notify the result periodly.

The current strategy for the filtering part is to start a new maple process. If the previous maple process is not finished, just kill it.

```
When query changes:

if last_maple_filtering_process not finished:
   kill(last_maple_filtering_process)

start_new_maple_process_for_filtering_given_new_query()
```

I know it's not efficient. The better one coule be starting the maple process only once when user invokes clap, and then whenever user change the query, maple's filter is also able to change the query on the fly. Do you have any ideas?

ImmemorConsultrixContrarie:
> Do you have any ideas?

I dunno. If you can use FFI Rust functions from vimscript, then cool, we can do it. If not, welp, just optimize what you can and kill the maple if you need to do something.

Right now I want you to battle-test new `fulf::starter()` function, while I write unit-tests. Whether it works at all and how good and fast it works. I'll add an option to pass different matchers and algos into it later, but it could be used and tested even without it.

liuchengxu:
> If you can use FFI Rust functions from vimscript

vimscript can't, but we can use Python/Lua for this purpose! And that's also what I'm interested in. Neovim has a builtin luajit, plus the overhead of using Python in Neovim https://github.com/liuchengxu/vim-clap/commit/75eb77ec5b6c0263cbfb55a90865234a3eafbf3d#r36451422, I personally prefer to use Rust FFI via luajit. I found https://github.com/distil/rust_lua_ffi later, but it's not popular like https://github.com/PyO3/pyo3 and not mature, so I did not complete the work of calling Rust via lua.

Actually I haven't got a clue to test `fulf::starter()`, you are intergating the feature of rg into the new lib, right?

ImmemorConsultrixContrarie:
> 
> 
> > If you can use FFI Rust functions from vimscript
> 
> vimscript can't, but we can use Python/Lua for this purpose! And that's also what I'm interested in. Neovim has a builtin luajit, plus the overhead of using Python in Neovim [75eb77e#r36451422](https://github.com/liuchengxu/vim-clap/commit/75eb77ec5b6c0263cbfb55a90865234a3eafbf3d#r36451422), I personally prefer to use Rust FFI via luajit. I found https://github.com/distil/rust_lua_ffi later, but it's not popular like https://github.com/PyO3/pyo3 and not mature, so I did not complete the work of calling Rust via lua.
> 
> Actually I haven't got a clue to test `fulf::starter()`, you are intergating the feature of rg into the new lib, right?

fulf reexports `ignore::Walk` (and `ignore::WalkBuilder` if you need some different from default settings), that just needs the path of directory to start from. You feed the path there, and it gives you the iterator over every non-ignored file in that path and subfolders. You then feed such iterator to the `fulf::starter()`.

Then there's `Rules` struct, and each field is public and documented to make things easier. Tell me if something is unclear there.

And last one - printer function. It should take two arguments:

1. Reference to already sorted results - `&[(Box<str>, Score, Box<[usize]>)]`. Could be changed later, but right now it's just `i32`.
2. A number of total results that get past the matcher.

Then you just feed the iterator, rules and a function to print results, and it should work.

ImmemorConsultrixContrarie:
I tweaked docs a bit. Have you tried `cd /vim-clap/crates/fulf` followed with `cargo doc --open`?
@liuchengxu

liuchengxu:
Hmm, I have looked at the doc a bit, but I have not found a clear way to use it in clap as it looks like a different workflow, the old way is to feed the output of fd/rg command to the filter.

The `dyn_fuzzy_filter_and_rank` works great, I'll firstly intergrate this feature into clap.

ImmemorConsultrixContrarie:
> the old way is to feed the output of fd/rg command to the filter

And what is the info you give fd/rg? Directory to search in with optional ignore rules, or nothing at all and it gets directory with `std::env:current_dir()`?

liuchengxu:
>what is the info you give fd/rg?

I merely pass the command string, e.g, `fd --type f`, to build the `std::process::Command`, optionally specify the command working directory by calling `command.current_dir()`, if no explicit working directory specified, use `std::env::current_dir()` then. 

I realize that with the new dynamic filter, if we can build an iterator of output from the command string, then we might almost reach the final goal, possibly like this example in rust-subprocess https://github.com/hniksic/rust-subprocess/blob/master/examples/sample.rs.

Current flow without dyn filter support:

1. Build `std::process::Command` C from string, e.g., `fd --type f`.
2. Wait for the complete output of C.
3. Feed the output of C into our dyn filter.

New flow with dyn filter support:

1. Build an iterator from the string, e.g., `fd --type f`.
2. Feed the iterator above into our dyn filter.

ImmemorConsultrixContrarie:
> 
> 
> > what is the info you give fd/rg?
> 
> I merely pass the command string, e.g, `fd --type f`, to build the `std::process::Command`, optionally specify the command working directory by calling `command.current_dir()`, if no explicit working directory specified, use `std::env::current_dir()` then.
> 
> I realize that with the new dynamic filter, if we can build an iterator of output from the command string, then we might almost reach the final goal, possibly like this example in rust-subprocess https://github.com/hniksic/rust-subprocess/blob/master/examples/sample.rs.
> 
> Current flow without dyn filter support:
> 
>     1. Build `std::process::Command` C from string, e.g., `fd --type f`.
> 
>     2. Wait for the complete output of C.
> 
>     3. Feed the output of C into our dyn filter.
> 
> 
> New flow with dyn filter support:
> 
>     1. Build an iterator from the string, e.g., `fd --type f`.
> 
>     2. Feed the iterator above into our dyn filter.

Just all lines from all files? Really? @liuchengxu

liuchengxu:
@ImmemorConsultrixContrarie  Hmm, I didn't make myself more clear, not all the lines of all files. For `grep`, I simply dispatch the rg command given the query and returns the raw output of rg, so it won't be all the lines of all files. But for `fd` to list the files, I do save all the lines it produces, it's accpetable since that normally has at most a few millions lines.

I believe fzf and skim both use the stream for reading data source and then do the filtering on them, which is missing in clap sadly at present :(.

crates/fulf/src/renamemeplsmynamesucks.rs
```diff
@@ -0,0 +1,342 @@
...
+    (shared, total)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::time::{Duration, SystemTime};
+
+    #[test]
+    fn basic_functionality_test() {
```
ImmemorConsultrixContrarie: Okay, I reworked that thing, now it's a bit different, and should work, because this basic test works.
I tested dynamic results on Rust src, and it works too.
Now I just need you to test the usability and speed of this new thing.

A bit or renaming will be done later, and there's no docs right now, but it works and could be tested.
Look in the test to get the idea of what `sort_and_print` function should look like and what to pass into (poorly named) `very_simple` function.

liuchengxu: I have to admit that I have not understand this new lib, which part it can replace and how it can be used in current flow. I'll spend more time after I merge your dyn filter feature into clap.

ImmemorConsultrixContrarie: > 
> 
> I have to admit that I have not understand this new lib, which part it can replace and how it can be used in current flow. I'll spend more time after I merge your dyn filter feature into clap.

Welp, everything, I guess. It uses the same recursive directory searcher as fd/rg, and when it finds a file, it reads the file, then iterates over lines of that file and applies matcher and scorer onto that line. When file ends, it searches for the next file.
With the last commit I changed `setter` function, so now it can take any algorithm.

ImmemorConsultrixContrarie:
Oh, wait. This is used to find files, not text in the files, yeah? Ahahahahahahaahahahahahahahahah. Sorry, that's nervous laughter.

@liuchengxu 
Please, tell me, that I didn't end up with thing that fuzzy-searches matches in files instead of files' names and pathes and even though it's finely multithreaded and should work fast, it's useless.

liuchengxu:
@ImmemorConsultrixContrarie It's totally ok for searching filenames. I'm nearly done with #364, then I can take time to interagte your new lib.

ImmemorConsultrixContrarie:
> 
> 
> @ImmemorConsultrixContrarie It's totally ok for searching filenames. I'm nearly done with #364, then I can take time to interagte your new lib.

No, it's not. Filename searching is a whole other thing, and it's almost impossible to get multithreaded filename searcher with *dynamic* results . FFFFFFFFFFFFFFFFUUUUUUUUUU...

To get real improvement in searching filenames, the algorithm should be changed, because right now it re-scores the whole path each time, even though many things share the same directory and could get some prescore.

And directory iterator should be changed if you want fzf-like search syntax, to stop searching if the directory got some `!fire` in it. Of course, you can just keep going in the directory, but what's the point if all results will be discarded?

Funny enough: the first thing I've done (wasn't even trying to get something cool, lol) was just what you needed, and all things I did after were useless. Hehe. Hehehe.

liuchengxu:
@ImmemorConsultrixContrari  🎉 Your dyn filter works in clap now! Really apprecicate your work.

I do agree with that the sorting/filtering approach for file names can be improved due to the current fuzzy filter algorithems are general, it's not optimal for the file names filtering. Taking the job of fd/rg in clap does not make much sense as they already the best tools for the time being IMHO :(.

The next optimization for clap would be making use of the cache of command that has a monster of results, e.g., 'cd / && fd --type f', find the files under the root directory, we can reuse the cached results for such jobs. That's my next move. fzf-like search syntax is a long-term goal, I actually haven't had a clear idea about that..

ImmemorConsultrixContrarie:
> Update: Just tried and it doesn't help much :( , decrease `ITEMS_TO_SHOW` is the most effective way, the current value is 100, I'll decrease it to 30 or even smaller.

I can see three ways: less number of things to print, bigger delay between prints, or just don't print anything at all if top results didn't change.

ImmemorConsultrixContrarie:
Hello, @liuchengxu !
I've had some rough times recently (I'm on linux now 'cuz my Windows 7 died with all the files it had, and a bunch of other problems), but now I'm back, probably.

I skimmed through the changelog, but haven't read the code yet, and to get into it fast I'll just ask some questions right there, okay?

So, the first question is about cache-feature. What do you cache actually? And how do you deal with ever-changing file tree (new folders, new files, deleted or moved things)?

I have some ideas about caching paths, as that will help computing next queries without going top-down through the whole system tree again, if invoked in the same folder or its child folder, and some other optimizations. But there are some really rough edges due to ever-changing nature of the file tree, thus I guess you are caching some other thing instead of paths?

The second question is simple: do you still want fzf-like syntax? I haven't seen anything about this in changelog.

liuchengxu:
@ImmemorConsultrixContrarie Definitely okay to ask here!

Let me answer the first cache question. I actually write the entire stdout of some command to a file when the stdout exceeds the threshold https://github.com/liuchengxu/vim-clap/blob/2d761834ef/crates/maple_cli/src/light_command.rs#L128. Then next time I could read from that cache file and no need to run the command again. The cache file is created by the formula `vim.clap/cmd_string/cmd_working_dir/cache_file`. For the same command running in the same directory, which happens a lot in vim-clap, this strategy just works.

With ever-changing file tree? No, I don't do that, I just provide a way for user to refresh the cache manually :(.

For the second question, yes, I still want fzf-like searching syntax, but I want to work out another problem first. Currently for each query string, maple process will be killed and rerun for the new query. I want to keep maple running for the whole session using the stdio-based RPC. The RPC framework is ready, the problem is to I need to access the whole result list during the session as I have to rerun the filtering job on the result list when new query comes, it's not hard if I already have the whole result list, I can just keep them in memory and access them at any time. But for some command that takes a while to complete, the whole result list is still a streamed stdout, I was wondering if it's good to write the stdout to a file when command is still streaming. But feel free to start ahead if you already have some ideas about fzf-like searching syntax.

PR #363, opened 2020-03-27
Use static reference to the const icon table

liuchengxu:
Ref https://github.com/liuchengxu/vim-clap/pull/353#issuecomment-604283141, cc @ImmemorConsultrixContrarie


ImmemorConsultrixContrarie:
Actually, I don't think that doing *everything* like it's done in stdlib is not right, as lowercase table is a bit different from this particular usecase.
For example, you lose some space there, as you can see [in those sizeofs](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9505c8f3c7e71ed9cf896f6ad48a9e85). Also, changing `&str` into `&[u8]` won't make any improvements, as those two things are just the same if all characters in the string are ASCII, but if you'd need a string out of this `[u8]` slice, you should write unsafe or lose time on checks.

liuchengxu:
@ImmemorConsultrixContrarie So, are you suggesting just using `&str` instead of `&[u8]` in our case?

ImmemorConsultrixContrarie:
> 
> 
> @ImmemorConsultrixContrarie So, are you suggesting just using `&str` instead of `&[u8]` in our case?

Yes. If you need a `&[u8]` out of a string, there's costless `&str.as_bytes()` method.

liuchengxu:
@ImmemorConsultrixContrarie Done.

PR #396, opened 2020-04-15
Introduce provider action using confirm()

### LINKED ISSUE #51: Support for buffer deletion through Clap buffer
When we have too many opened buffers in a vim session, it will be nice to have feature to mark and delete buffers not required without restarting vim.

liuchengxu:
Close #51

The action dialog is implemented using `confirm()` for neovim, popup for vim.

Vim:

截屏2020-04-19 上午1 07 36

NeoVim:

截屏2020-04-19 上午1 10 59


vn-ki:
Is it not possible to use popup for nvim?

liuchengxu:
@vn-ki Definitely possible, but need some time to implement `popup_menu` like component using neovim's floating_win.

PR #405, opened 2020-04-18
Fix binary installation on Windows

### LINKED ISSUE #404: :Clap install-binary! Failed on windows 10
**Environment**
 - OS: windows 10
 - Neovim v0.4.3, Gvim v8.2
 - vim-clap version: v0.12
 - Have you reproduced with a minimal vimrc: Yes
 - Have you updated to the latest plugin version: Yes
 - Have you upgraded to/compiled the latest Rust binary:  Having issue when running `:Clap install-binary!`

**Describe the bug**

`:Clap install-binary!` failed on windows 10.

When the path to the install.ps1 contains spaces,  it will be considered as several separate parameters.  The fix is to quote the path to the install.ps1.

And even if the path doesn't contain spaces , windows 10 would prevent user running powershell script based on the default execution policy configuration. The fix is when executing powershell script, add ` -ExecutionPolicy ByPass` to be able to run it. 

Ref: [Powershell Execution Policies](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7)

And even if all the above two problems are fixed, the install.ps1 will still fail showing the DownloadFile function cannot create SSL/TLS secure channel. This is because TLS12 is not enabled in powershell by default. The fix is to add TLS12 to the security protocol .
Mentioned by @masaeedu in https://github.com/lukesampson/scoop/pull/2065#issuecomment-369669048
```powershell
[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
```
[Net.ServicePointManager](https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.securityprotocol?view=netframework-4.8)
[SecurityProtocolType Enum](https://docs.microsoft.com/en-us/dotnet/api/system.net.securityprotocoltype?view=netframework-4.8#System_Net_SecurityProtocolType_SystemDefault)

**To Reproduce**

1. Create the minimal vimrc `min.vim`:

```vim
set nocompatible
set runtimepath^=/path/to/vim-clap
syntax on
filetype plugin indent on
```
Case 1: path contains space
Case 2: path is good
Case 3: path is good and no issues with execution policy

2. Start (neo)vim with command: `vim -u min.vim`

3. Type `:Clap install-binary!`
Case 1
IEUPX``2(BWZD(NT9UUM6JQ
Case 2
}%@`R{GDPS55_SF@NDAPOOD
Actually this is an output from the command line, but you will see same output from the vim.
Case 3
D MT46 (Z2E9T$TZ 0X) {0

**Expected behavior**
Install binaries successfully and see the success message.

I created a PR which did the following things:
1. quote the path to the install.ps1
2. bypass powershell's execution policy
3. Add Tls12 to the security protocal

I tested for both Neovim v0.4.3 and Gvim v8.2 on windows10.
Created a pull request for this #405 . What do you think?

TissueFluid:
Fix #404 

- Quote the path to install.ps1 so that it will handle spaces correctly.
- Add `-ExecutionPolicy ByPass`  as a parameter to powershell to be able to execute powershell script when by default win10 prevent you to do this.
- Add Tls12 to security protocol to be able to download the pre-built binary.

### DIFF
--- a/autoload/clap/installer.vim
+++ b/autoload/clap/installer.vim
@@ -7,7 +7,7 @@ set cpoptions&vim
 let s:plugin_root_dir = fnamemodify(g:clap#autoload_dir, ':h')
 
 function! s:run_term(cmd, cwd, success_info) abort
-  10new belowright bottom
+  belowright 10new bottom
   setlocal buftype=nofile winfixheight norelativenumber nonumber bufhidden=wipe
 
   function! s:OnExit(status) closure abort
@@ -106,7 +106,7 @@ endfunction
 
 function! clap#installer#download_binary() abort
   if has('win32')
-    let cmd = 'Powershell.exe -File '.s:plugin_root_dir.'\install.ps1'
+    let cmd = 'Powershell.exe -ExecutionPolicy ByPass -File "'.s:plugin_root_dir.'\install.ps1"'
   else
     let cmd = './install.sh'
   endif
--- a/autoload/clap/installer.vim
+++ b/autoload/clap/installer.vim
@@ -7,8 +7,10 @@ set cpoptions&vim
 let s:plugin_root_dir = fnamemodify(g:clap#autoload_dir, ':h')
 
 function! s:run_term(cmd, cwd, success_info) abort
-  belowright 10new bottom
+  belowright 10new
   setlocal buftype=nofile winfixheight norelativenumber nonumber bufhidden=wipe
+
+  let bufnr = bufnr('')
 
   function! s:OnExit(status) closure abort
     if a:status == 0
@@ -29,8 +31,6 @@ function! s:run_term(cmd, cwd, success_info) abort
           \ 'exit_cb': {job, status -> s:OnExit(status)},
           \})
   endif
-
-  let bufnr = bufnr('')
 
   normal! G
 
--- a/autoload/clap/maple.vim
+++ b/autoload/clap/maple.vim
@@ -16,10 +16,10 @@ let s:maple_bin_prebuilt = fnamemodify(g:clap#autoload_dir, ':h').'/bin/maple'.s
 
 " Check the local built.
 if executable(s:maple_bin_localbuilt)
-  let s:maple_bin = s:maple_bin_localbuilt
+  let s:maple_bin = '"' . s:maple_bin_localbuilt . '"'
 " Check the prebuilt binary.
 elseif executable(s:maple_bin_prebuilt)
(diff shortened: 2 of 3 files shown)


autoload/clap/installer.vim
```diff
@@ -7,7 +7,7 @@ set cpoptions&vim
 let s:plugin_root_dir = fnamemodify(g:clap#autoload_dir, ':h')
 
 function! s:run_term(cmd, cwd, success_info) abort
-  10new belowright bottom
+  belowright 10new bottom
   setlocal buftype=nofile winfixheight norelativenumber nonumber bufhidden=wipe
```
TissueFluid, line 11: Is this one a typo? Because `10new belowright` will actually create a window whose name is "belowright bottom".

TissueFluid:
Found some other issues. Discussed in #405

autoload/clap/installer.vim
```diff
@@ -7,9 +7,11 @@ set cpoptions&vim
 let s:plugin_root_dir = fnamemodify(g:clap#autoload_dir, ':h')
 
 function! s:run_term(cmd, cwd, success_info) abort
-  belowright 10new bottom
+  belowright 10new
```
TissueFluid, line 10: When user enabled swapfile, this line will have an error, saying trying to rename a swap file. So I just removed  the name of this window.

autoload/clap/installer.vim
```diff
@@ -30,8 +32,6 @@ function! s:run_term(cmd, cwd, success_info) abort
           \})
   endif
 
-  let bufnr = bufnr('')
-
   normal! G
```
TissueFluid, line 35: This line should be placed before bufnr is used?

liuchengxu: Not sure, the current one works for me, both vim and neovim. If this change is helpful for you, I'm ok to include.

autoload/clap/installer.vim
liuchengxu: Ok to me.

liuchengxu:
I think is this ready to merge?

TissueFluid:
> I think is this ready to merge?

Just pushed the change to fix the 👏 grep2. 1 sec.

TissueFluid:
All checks have passed. Feel free to merge. Thanks.

PR #408, opened 2020-04-22
feat(filer): jump directly up to parent directory

michaelbeaumont:
Loving this plugin so far, but one thing that slows me down a bit is when I'm using `filer` and I want to jump to the parent but I've typed some characters already, so I have to press `BS` a bunch first to get rid of the characters. I *think* `ivy` had a feature where you could directly jump to the parent and I *think* it was bound to `<S-Tab>`. (I might be imagining this)

This PR introduces that feature. I know `<S-Tab>` is bound to `invoke` in insert mode, which would be confusing in this case, so probably the actual keybinding needs to change.

What do you think?

### DIFF
--- a/autoload/clap/handler.vim
+++ b/autoload/clap/handler.vim
@@ -101,6 +101,13 @@ function! clap#handler#tab_action() abort
   return clap#selection#toggle()
 endfunction
 
+function! clap#handler#stab_action() abort
+  if has_key(g:clap.provider._(), 'stab_action')
+    call g:clap.provider._().stab_action()
+  endif
+  return ''
+endfunction
+
 function! clap#handler#bs_action() abort
   if has_key(g:clap.provider._(), 'bs_action')
     call g:clap.provider._().bs_action()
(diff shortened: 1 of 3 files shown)


liuchengxu:
I'm ok with using a new key binding for this feature, but you know it's not easy to come up with a new unused key binding that mostly works out of box.

michaelbeaumont:
I suppose changing the keybinding for `invoke` is out of the question :grimacing:. For my use case it'd be enough to just offer the action itself (and rename it to something like `alt_action`) without a default keybinding.

liuchengxu:
`<M-*>` does not always work out of box, so it's not suitable to be the default key binding.

liuchengxu:
To meet your requirement, actually what you all need to do is to add these lines in your vimrc

```vim
" For filer provider, neovim only
function! MyShiftTab() abort
  call g:clap.input.set('')
  call g:clap.provider._().bs_action()
  return ''
endfunction

autocmd FileType clap_input inoremap <silent> <buffer> <S-Tab> <C-R>=MyShiftTab()<CR>
```

michaelbeaumont:
very true, if you're not sure if others would appreciate the functionality, that's enough for me!

liuchengxu:
I always try to keep Vim and NeoVim work in a same way, so I'm reluctant to accept the change that works for neovim only. Thanks anyway!

PR #417, opened 2020-04-29
Show new file as explicit entry in filer

michaelbeaumont:
This makes filer always shows an entry for the query itself in the file
list, which makes editing a new file more explicit.

Before, if the query didn't match anything "NO MATCHES FOUND" is shown
and if you press enter, clap opens a buffer named the current query.
It also wasn't possible to open a new buffer this way if the directory
was empty. A buffer with the name "Directory is empty" is opened no
matter what the input is.

Now, the current query is always shown as a file that can be opened,
which, at least for me, feels more intuitive and matches behavior of
similar interfaces (e.g. Helm find files in emacs).
This also works in empty directories. If you tab to an empty directory
and press enter with the empty query, the directory itself is opened.

### DIFF
--- a/autoload/clap/provider/filer.vim
+++ b/autoload/clap/provider/filer.vim
@@ -116,9 +116,9 @@ function! s:do_filter() abort
     call g:clap.display.set_lines(candidates)
     call g:clap#display_win.shrink_if_undersize()
   else
-    let query_entry = g:clap_enable_icon ? ' '.query : query
+    let query_prefix = '+ '
     call clap#filter#on_typed(
-      \ function('clap#filter#sync'), query, candidates + [query_entry]
+      \ function('clap#filter#sync'), query, candidates, query_prefix
     \ )
   endif
 endfunction
--- a/autoload/clap/provider/filer.vim
+++ b/autoload/clap/provider/filer.vim
@@ -8,13 +8,6 @@ set cpoptions&vim
 
 let s:filer = {}
 
-let s:DIRECTORY_IS_EMPTY = (g:clap_enable_icon ? '  ' : '').'Directory is empty'
-
-function! clap#provider#filer#hi_empty_dir() abort
-  syntax match ClapEmptyDirectory /^.*Directory is empty/
-  hi default link ClapEmptyDirectory WarningMsg
-endfunction
-
 function! s:handle_round_message(message) abort
   try
     let decoded = json_decode(a:message)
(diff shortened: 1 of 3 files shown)


liuchengxu:
>It also wasn't possible to open a new buffer this way if the directory was empty. 

To achieve this, only the `sink` function needs to be changed.

>If you tab to an empty directory and press enter with the empty query, the directory itself is opened.

Ditto. Only `sink` needs to modified. I don't understand why the other changes are neccessary? Can you explain a bit?

michaelbeaumont:
It isn't possible to add the query as an entry in the list just changing `sink`, which was the main reason I made the change, along with the "Directory is empty" bug.
The behavior until now is (`file` is the query):
```
/home/mike/folder/[file                                          ]
>>file1
>>file2
```
and I am unable to create `file` because it matches an existing entry.

Now:
```
/home/mike/folder/[file                                          ]
>>file1
>>file2
>>file
```
where `file` doesn't exist.

liuchengxu:
Your current approach doesn't make much sense semantically as the query is not a candidate essentially, especially injecting it in the candidate list when filtering. Have you check about the `action` feature introduced since https://github.com/liuchengxu/vim-clap/pull/396? It's a better way to work with these cases, not only creating a new file, but deleting and moving, etc.

liuchengxu:
>it seems like they operate on an entry in the list

That's incorrect. You can do basically anything as you can access all the information.

I can accept you also implement the same thing like helm, using a special mark for that mocked entry makes more sense.

michaelbeaumont:
@liuchengxu ok, I've added the prefix (I used `+ `)
I also had to fix the entry appearing double if the file _does_ exist, it's a bit little bit ugly, though. This was the easiest way, although it might be worth building into the rust crate.

autoload/clap/provider/filer.vim
```diff
@@ -36,13 +29,8 @@ function! s:handle_round_message(message) abort
...
     let result = decoded.result
-    if result.total == 0
-      let s:filer_empty_cache[result.dir] = s:DIRECTORY_IS_EMPTY
-      call g:clap.display.set_lines([s:DIRECTORY_IS_EMPTY])
-    else
-      let s:filer_cache[result.dir] = result.entries
-      call g:clap.display.set_lines(result.entries)
-    endif
+    let s:filer_cache[result.dir] = result.entries
+    call g:clap.display.set_lines(result.entries)
```
liuchengxu, line 33: You should inject the `query` candidate here, and `clap#filter#on_typed()` don't have to be changed.

autoload/clap/provider/filer.vim
```diff
@@ -128,7 +116,10 @@ function! s:do_filter() abort
     call g:clap.display.set_lines(candidates)
     call g:clap#display_win.shrink_if_undersize()
   else
-    call clap#filter#on_typed(function('clap#filter#sync'), query, candidates)
+    let query_prefix = '+ '
+    call clap#filter#on_typed(
+      \ function('clap#filter#sync'), query, candidates, query_prefix
+    \ )
```
liuchengxu, line 122:
```vim
call clap#filter#on_typed(function('clap#filter#sync'), query, candidates[1:])
```

autoload/clap/provider/filer.vim
```diff
@@ -173,22 +164,16 @@ function! s:tab_action() abort
     return
   endif
 
-  if has_key(s:filer_empty_cache, s:current_dir)
-    if g:clap.display.get_lines() != [s:DIRECTORY_IS_EMPTY]
-      call g:clap.display.set_lines([s:DIRECTORY_IS_EMPTY])
```
liuchengxu, line 178: I think it's still useful to indicate that directory is empty even you inject the faked query candidate.

liuchengxu:
To implement the helm like new file entry, a lot of details needs to be taken care of.

michaelbeaumont:
I've been using my changes on top of vim-clap for the past 2 months w/o issue, so I haven't bothered to update this. I'll take another look.

PR #423, opened 2020-05-03
FUUUUUUUUUUULF!

ImmemorConsultrixContrarie:
Okay, let's start this over again.
I tried to make the behaviour identical to that rg command, but I'm not sure, if it's really identical.

P.S. Sorry for that PR name.


crates/fulf/src/interface/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+    ///
+    /// 1. This is a line in a text that is not code.
+    ///
+    /// 2. This is a non-formatted line of automatically generated code.
+    ///
+    /// 3. This is a very bad code.
+    ///
+    /// 4. Some very rare other reasons, like giant right-shifted branching.
+    ///
+    /// And in any of those cases there's probably no point in fuzzing such line.
```
ImmemorConsultrixContrarie, line 519: Some explaines in those comments.
Btw, right now this `max_line_len` is used under the hood:
https://github.com/liuchengxu/vim-clap/blob/60811e3dcb5b5f7789985e9955eb47049f89c51d/crates/extracted_fzy/src/lib.rs#L65

I'm just using it explicitly and more effectively, 'cuz I check the length before matching anything in the lines.

crates/fulf/src/interface/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+                        if let Some(pack) = results.get(idx) {
+                            let (s, _score, pos) = pack;
+                            println!("Total: {}\n{}\n{:?}", total, s, pos);
+                        } else {
+                            break;
+                        }
+                    }
+                }
+            }
+        };
```
ImmemorConsultrixContrarie, line 602: This part in the test shows the signature of the required `sort_and_print` function.

But due to the multithreading of the whole thing, it's different from the one-threaded dynamic filter; there won't be as many prints as there is in one-threaded filter due to the overhead of synchronization of many threads. I will improve the synchronization routine soon, but the main problem will still be here: you can only get results when the filter produces `results_cap` items. Probably, this filter should beat the rarity of updates with better general speed.

crates/fulf/src/interface/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+                0
+            };
+
+            r
+        };
+
+        let path = path.as_ref();
+        let dir_iter = ignore::Walk::new(path);
+
+        let root_folder = path.to_str()?;
```
ImmemorConsultrixContrarie, line 549: In that line the whole thing could fail because:

First: if you use `OsStr::to_string_lossy()`, you cannot restore original path.
Second: there is no easy way to represent the path as raw bytes.

Thus it's easier to ignore any path that can't be represented as utf8. Anyway, no sane person would store anything useful in a path that could not be represented as Unicode.

crates/fulf/src/interface/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+            FzyAscii::setter(dir_iter, fzydata, r, sort_and_print)
+        } else {
+            // utf8
+            let algo = move |needle: &str, line: &str| {
+                if line.len() > max_line_len {
+                    None
+                } else {
+                    utf8::match_and_score_with_positions(needle, line)
+                }
+            };
```
ImmemorConsultrixContrarie, line 565: Oops! There's a mistake: I made the the rg-like lines, but completely forgot to drop the filename from the scored string. I'll fix it as soon as possible.

Btw, I can just give the length of the partial path right to the algorithm, so there's no need in regex evaluation. Also, I can trim the whitespaces from the start of the line and add the trimmed number to the column number. Probably, should not trim all whitespaces, because it would change the picture of a line?
I dunno whether trimming all the spaces from the start or leaving up to 2-4 spaces would be better, need other opinion there.

liuchengxu: > can trim the whitespaces from the start of the line and add the trimmed number to the column number.

Actually I want this before! The leading whitespaces are meaningless, I think it's fine to trim the leading and trailing whitespaces.

crates/fulf/src/ascii/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+
+mod bytelines;
+pub use bytelines::ByteLines;
+mod matcher;
+pub use matcher::{ascii_from_bytes, matcher};
+
+use crate::scoring_utils::*;
+
+#[inline]
+pub fn score_with_positions(needle: &[u8], haystack: &[u8]) -> (Score, Vec<usize>) {
```
liuchengxu, line 13: hmm, why we are duplicating this here?

crates/fulf/src/interface/mod.rs
liuchengxu: We could make `results_cap` be 128, or even 64, that's totally fine.

crates/fulf/src/ascii/mod.rs
ImmemorConsultrixContrarie: > hmm, why we are duplicating this here?

It was easier to change the code in just one crate, plus, `ascii` module is a bit more than just fzy algorithm, if you are asking why I duplicate this with "extracted_fzy". Of course, those two crates should not duplicate the code, but I'm not sure how they should be merged. Probably, algorithm-specific things should be moved out of the `fulf`.

Or, if you are asking why there is "utf8" and "ascii" with duplicated code - that's because they work with different types: ascii algorithm works with byteslices and bytes, where utf8 algorithm works with string and chars.

crates/fulf/src/interface/mod.rs
ImmemorConsultrixContrarie: > We could make `results_cap` be 128, or even 64, that's totally fine.

Yes... but at the same time, no. Less results to store - blocking thread synchronization is done more often. Actually, 128 or 64 should work almost as good as 512, but I'm not sure about the real cost.

crates/fulf/src/ascii/mod.rs
liuchengxu: To be merged into the master, we better at least refactor the duplicated `extracted_fzy` code in case of the bug of `extracted_fzy` like https://github.com/liuchengxu/vim-clap/pull/409/files#diff-744145d5678fb6819a7b4fddd9b56248R94 once we finish the other works in this PR.

ImmemorConsultrixContrarie: > To be merged into the master, we better at least refactor the duplicated `extracted_fzy` code in case of the bug of `extracted_fzy` like https://github.com/liuchengxu/vim-clap/pull/409/files#diff-744145d5678fb6819a7b4fddd9b56248R94 once we finish the other works in this PR.

I copied the fixed algorithm, and added the ASCII there, watch c7a7a94 commit. Should be fine.

crates/fulf/src/interface/mod.rs
ImmemorConsultrixContrarie: > > can trim the whitespaces from the start of the line and add the trimmed number to the column number.
> 
> Actually I want this before! The leading whitespaces are meaningless, I think it's fine to trim the leading and trailing whitespaces.

It's there: aeabfe1.

crates/fulf/src/interface/mod.rs
liuchengxu: Haven't really used it, so I'm also not sure if 512 or 128 has a different effect noticeably.

ImmemorConsultrixContrarie:
Looks like that filepath cache works for ASCII (tested for the whole "/home" directory), but more tests on non-ASCII UTF-8 paths should be done.

liuchengxu:
@ImmemorConsultrixContrarie Sorry for taking such a long time, I have just resolved the conflict and start to take a look at it again, hopefully we could make some progress this time. Could you make the function of `show_case` module a public usable API rather than a test feature? I think it's ready to be used directly. For the library user, I prefer an API the default `handle_results`, I only need to pass the `path`, `needle` and `max_line_len`, the default `handle_results` could just print the results to stdout, that's fine.

ImmemorConsultrixContrarie:
> @ImmemorConsultrixContrarie Sorry for taking such a long time, I have just resolved the conflict and start to take a look at it again, hopefully we could make some progress this time. Could you make the function of `show_case` module a public usable API rather than a test feature? I think it's ready to be used directly. For the library user, I prefer an API with an optional default `handle_results`, I only need to pass the `path`, `needle` and `max_line_len`, the default `handle_results` could just print the results to stdout, that's fine. If we could add a CLI interface for fulf in maple_cli, that's even better.

Well, I don't really like the code here after those four months. Over this time I learned a little about IPC and overall application architecture, and can see some flaws here.

First, the main speed problem wasn't solved: starting and killing the binary each time the input changes is really slow. Instead it should be a process that lives as long as [vim](https://vimhelp.org/channel.txt.html) or [nvim](https://neovim.io/doc/user/channel.html) instance lives itself.

Second, the cache. It should store key-value map for each command (or just some of the last used commands) where `key` is a file, and `value` is a list of results. Then, when a [file changes](https://docs.rs/notify/4.0.15/notify/), the cache for the changed file is marked as invalid. Next time the command is evoked we just read valid results and recalculate invalid results, then merge those.
Although there could be some problems with multiple instances of the application.

Third, some little things I don't like when I see them, like reading the whole file instead of using `BufReader`.

So, the answer to your question is "No I can't, because I want to rewrite this thing completely".

liuchengxu:
>I don't like JSON for performance reasons, so wanna rewrite this too.

@ImmemorConsultrixContrarie  Please just go for it! I'm personally not familiar with IPC, but always hope to find a chance to try it out, would be super cool to use IPC in vim-clap 👍

ImmemorConsultrixContrarie:
> @ImmemorConsultrixContrarie Please just go for it! I'm personally not familiar with IPC, but always hope to find a chance to try it out, would be super cool to use IPC in vim-clap +1

IPC is just a short name for [inter-process communication](https://en.wikipedia.org/wiki/Inter-process_communication) and stdin-stdout pipe is one of those. Thus yeah, you are already using it.

liuchengxu:
@ImmemorConsultrixContrarie Good to know stdio is also one of IPC stuffs.

liuchengxu:
@ImmemorConsultrixContrarie This is continued in #875 to some extent after 2 years, feel free to leave any comments there if it still looks interesting :(

liuchengxu:
Thanks for the efforts in this PR, it now basically has been implemented on the master branch [using the library ripgrep](https://github.com/liuchengxu/vim-clap/blob/0571d4487a/crates/maple_core/src/searcher/grep.rs). Again, really appreciate all the contributions you've made!

PR #424, opened 2020-05-04
Global variable :: clap_preview_files_lines

lcrockett:
# Description
When using `:Clap files` allow number of lines shown in preview to be customisable using a global variable `clap_preview_files_lines`.

# Documentation proposal
```
g:clap_preview_files_lines                              *g:clap_preview_files_lines*

  Type: |Number|
  Default: `10`

  This variable control the number of lines previewed using :Clap files
```

### DIFF
--- a/autoload/clap/preview.vim
+++ b/autoload/clap/preview.vim
@@ -7,7 +7,7 @@ set cpoptions&vim
 let s:path_seperator = has('win32') ? '\' : '/'
 
 function! s:peek_file(fname, fpath) abort
-  let lines = readfile(a:fpath, '', 10)
+  let lines = readfile(a:fpath, '', get(g:, 'clap_preview_files_lines', '10'))
   call insert(lines, a:fpath)
   call g:clap.preview.show(lines)
   call g:clap.preview.set_syntax(clap#ext#into_filetype(a:fname))


autoload/clap/preview.vim
```diff
@@ -7,7 +7,7 @@ set cpoptions&vim
 let s:path_seperator = has('win32') ? '\' : '/'
 
 function! s:peek_file(fname, fpath) abort
-  let lines = readfile(a:fpath, '', 10)
+  let lines = readfile(a:fpath, '', get(g:, 'clap_preview_files_lines', '10'))
```
liuchengxu, line 10: If we choose to use a global variable to control the number of preview lines, I hope it's valid for all the providers, not   for `files` like provider only. The problem is currently the preview impl is relatively scattered, e.g., `grep` also has its own choice https://github.com/liuchengxu/vim-clap/blob/60811e3dcb5b5f7789985e9955eb47049f89c51d/autoload/clap/provider/grep.vim#L213 .

lcrockett: You're a minute faster than me; I just created a second PR dealing with `:Clap grep`. Subsequently i've proposed a global variable name that includes the provider reference to keep behaviour separated and changeable per provider. `clap_preview_files_lines` + `clap_preview_grep_lines`.

liuchengxu: Hmm, I mean a global variable for all the providers, not one per provider. I'm not a fun of too many options that are not neccessary.

lcrockett: I think it only makes sense to allow for manipulation of preview lines with providers `git_files`, `filer`, `files`, and `grep`.

However, 4 different variables seems a bit excessive to me too. I do however see a useful use-case in creating 2 separate variables dealing with preview lines: 1 for `git_files + filer + files` and 1 for `grep`.

It is plausible that when using the `grep` provider you want fewer previewed lines as opposed to using the `git_files / filer / files` providers as you kinda have an idea what you're looking for when using `grep` and in the case of the latter you might want to have a few more previewed lines to increase visibility.

At least in my case as i'd set `files` to 20 lines and `grep` to 15 lines.

How about 2 variables
* clap_preview_files_lines :: Affecting `git_files` + `filer` + `files`
* clap_preview_grep_lines :: Affecting `grep`

liuchengxu: >I think it only makes sense to allow for manipulation of preview lines with providers git_files, filer, files, and grep.

I don't think so, `tags`, `btags`, `marks`, `buffers`, `blines` providers all use the same `file` like preview impl.

We can use a global variable, which could be a Number or a Dict. If it's a Number, all providers use the same preview size. If it's a Dict, each provider can specify its own preferred size.

lcrockett: Ah, I wasn't aware that those providers also use the same preview impl. Managing behaviour based on usage of `Dict` or `Number` seems like a good compromise that allows both use-cases. I would say the variable name could be along the lines of `clap_preview_lines` emphasising it is not bound to a specific provider.

As this proposal makes my PR for `clap_preview_grep_lines` obsolete, i'll close that one.

liuchengxu:
@lcrockett Try this branch https://github.com/liuchengxu/vim-clap/tree/clap-preview-size, `:h g:clap_preview_size` for the usage.

lcrockett:
Can confirm it works well, `let g:clap_preview_size = { '*': 7, 'files':12, 'grep': 10 }`. Cheers for the merge and changes !

Issue #450, opened 2020-06-01
Some of the text is hard to identify in dark theme


ray-x:
This is more an improvement suggestion instead of a Bug.

**Environment (please complete the following information):**
 - OS: MacOS
 - (Neo)Vim version: 8.2
 - vim-clap version: master
 - Have you reproduced with a minimal vimrc: yes
 - Have you updated to the latest plugin version: yes
 - Have you upgraded to/compiled the latest Rust binary: yes 

**Describe the bug**

Please refer to the attached images. 
When I use Clap proj_tags, the `tag` field is hard to identify with (any) dark theme.

**Screenshots**
Please check the texts I marked with red
![clap_tags](https://user-images.githubusercontent.com/1681295/83379881-eb7cfd80-a41f-11ea-938c-76ebd2878b41.jpg)

liuchengxu:
The default highlight can't be guaranteed to work with all themes, refer to https://github.com/liuchengxu/vim-clap/blob/master/syntax/clap_proj_tags.vim to custom these highlights.

For me, I use https://github.com/liuchengxu/space-vim-theme:

<img width="1142" alt="截屏2020-06-01 下午3 34 21" src="https://user-images.githubusercontent.com/8850248/83386283-5f69d680-a41d-11ea-9d50-955af2ece952.png">

ray-x:
Thanks for the suggestions,

One issue that all the default themes provided by clap (atom_dark,  material_design_dark, Nord, solarized_dark) has same problem. So it took me some time to try the provided theme but none of them work.

The reason behind is clap float window bg color is gray and specialKey(which is used for show main body of proj_tag) is also gray in most dark theme.

I would recommend either change the theme provided by clap (may be add space-vim as one of the default theme). Or change `ClapProjTagPattern` color to `String` (or what ever that are not `specialKey`)

liuchengxu:
>Or change ClapProjTagPattern color to String (or what ever that are not specialKey)

Feel free to send a PR.

PR #451, opened 2020-06-01
Update the highlight for proj_tags #450

ray-x:
Improvement for 
 ['Some of the text is hard to identify in dark theme #450'](https://github.com/liuchengxu/vim-clap/issues/450)
One liner.

### DIFF
--- a/syntax/clap_proj_tags.vim
+++ b/syntax/clap_proj_tags.vim
@@ -9,7 +9,7 @@ syntax match ClapProjTagPattern /^.*$/ contains=ClapTagName,ClapProjTagKind,Clap
 hi default link ClapProjTagName Type
 hi default link ClapProjTagKind Function
 hi default link ClapProjTagPath Directory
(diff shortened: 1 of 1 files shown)


syntax/clap_proj_tags.vim
```diff
@@ -9,7 +9,7 @@ syntax match ClapProjTagPattern /^.*$/ contains=ClapTagName,ClapProjTagKind,Clap
 hi default link ClapProjTagName Type
 hi default link ClapProjTagKind Function
 hi default link ClapProjTagPath Directory
-hi default link ClapProjTagPattern SpecialKey
+hi default link ClapProjTagPattern String
```
liuchengxu, line 12: I suggest using another one as `ClapProjTagKindPathSeperator` already uses String.

ray-x: Agree, changed to `Identifier`

Issue #454, opened 2020-06-04
Preview window overlap with file list window in neovim-0.5


ray-x:
**Environment (please complete the following information):**
 - OS: Mac
 - (Neo)Vim version: Neovim-0.5 (--HEAD)
 - vim-clap version: latest

**Describe the bug**
In neovim, float window seems to be flexible and  preview could overlap list window (please see attached image) The selected is hidden when  I move cursor down. 

**To Reproduce**
Install both neovim-0.4.3 and neovim-0.5.  Same configure but neovim 0.5 does not work correctly.

**Expected behavior**
neovim-0.5 should behave same as neovim-0.4.3

**Screenshots**
begin search and move cursor to last item in list:
![Screen Shot 2020-06-04 at 2 34 53 pm](https://user-images.githubusercontent.com/1681295/83716167-7c4b1780-a672-11ea-8b4b-0a86aab8876c.png)

continue move with `down`, cursor is gone from search list and I can see highlight move in preview list(highlighted):
![Screen Shot 2020-06-04 at 2 37 23 pm](https://user-images.githubusercontent.com/1681295/83716192-8cfb8d80-a672-11ea-8eb8-3da07c4edf43.png)

Continue move with down:
![Screen Shot 2020-06-04 at 2 37 32 pm](https://user-images.githubusercontent.com/1681295/83716200-92f16e80-a672-11ea-8741-8a994d60ef1a.png)

Also, The search result list window size seem to change with cursor movement, in previous screen shot, list is 3 items(fixed), in the screen shot, it grow to 12.
![Screen Shot 2020-06-04 at 2 36 27 pm](https://user-images.githubusercontent.com/1681295/83716217-9dac0380-a672-11ea-98fc-07ab02ac15ab.png)

**Additional context**

liuchengxu:
- I don't want to spend time debugging the unstable neovim version.
- vim-clap has a cache feature which will load the lines lazily. If you keep moving the cursor down, and the cache has more lines, these cached lines will be loaded then. `:Clap grep` has a new versin `:Clap grep2`, you can try that too.

ray-x:
Thanks for your suggestions, I will use grep2 more.
Also, for this defect, `:Clap grep2` also has similar issue.

liuchengxu:
I upgraded my neovim to HEAD-96568d5, still can't reproduce it.

ray-x:
[...]

PR #458, opened 2020-06-06
[#454]Fix search and preview overlap

ray-x:
Details https://github.com/liuchengxu/vim-clap/issues/454
recalculate the preview windows size.

### DIFF
--- a/autoload/clap/preview.vim
+++ b/autoload/clap/preview.vim
@@ -21,16 +21,29 @@ function! s:show_file_props(entry) abort
   call clap#preview#highlight_header()
 endfunction
 
+function! s:preview_lines_calc(h) abort
+  let l:layout = clap#layout#calc()
+  let l:remain = ( &lines - l:layout.row - l:layout.height - 1 ) / 2
+  let l:size = a:h
+  if l:remain <= 0
+    return 0
+  endif
+  if l:size <= l:remain
+    return l:size
+  endif
+  return l:remain
+endfunction
+
 if type(g:clap_preview_size) == v:t_number
   function! clap#preview#size_of(provider_id) abort
-    return g:clap_preview_size
+    return s:preview_lines_calc(g:clap_preview_size)
   endfunction
 elseif type(g:clap_preview_size) == v:t_dict
   function! clap#preview#size_of(provider_id) abort
     if has_key(g:clap_preview_size, a:provider_id)
-      return g:clap_preview_size[a:provider_id]
+      return s:preview_lines_calc(g:clap_preview_size[a:provider_id])
     elseif has_key(g:clap_preview_size, '*')
-      return g:clap_preview_size['*']
+      return s:preview_lines_calc(g:clap_preview_size['*'])
     else
       return s:default_size
     endif
@@ -93,10 +106,10 @@ endfunction
 " the origin line should be positioned.
 " 0-based
 function! clap#preview#get_line_range(origin_lnum, range_size) abort
-  if a:origin_lnum - a:range_size > 0
+  if a:origin_lnum > a:range_size
     return [a:origin_lnum - a:range_size, a:origin_lnum + a:range_size, a:range_size]
   else
-    return [0, a:origin_lnum + a:range_size, a:origin_lnum]
+    return [0, a:range_size * 2, a:range_size]
   endif
 endfunction
 


autoload/clap/preview.vim
```diff
@@ -93,10 +106,10 @@ endfunction
 " the origin line should be positioned.
 " 0-based
 function! clap#preview#get_line_range(origin_lnum, range_size) abort
-  if a:origin_lnum - a:range_size > 0
+  if a:origin_lnum > a:range_size
     return [a:origin_lnum - a:range_size, a:origin_lnum + a:range_size, a:range_size]
   else
-    return [0, a:origin_lnum + a:range_size, a:origin_lnum]
+    return [0, a:range_size * 2, a:range_size]
```
liuchengxu, line 112: What problem this change resolves?

autoload/clap/preview.vim
```diff
@@ -21,16 +21,29 @@ function! s:show_file_props(entry) abort
...
+  if l:size <= l:remain
+    return l:size
+  endif
+  return l:remain
+endfunction
+
 if type(g:clap_preview_size) == v:t_number
   function! clap#preview#size_of(provider_id) abort
-    return g:clap_preview_size
+    return s:preview_lines_calc(g:clap_preview_size)
```
liuchengxu, line 39: Don't modify `clap#preview#size_of()` directly which is used to get the raw user defined setting, define a new function for that.

liuchengxu: I suggest tweaking the actual final preview size in `clap#floating_win#preview.show()`, https://github.com/liuchengxu/vim-clap/blob/2ad4ce34a143b7734577388fba765d637f64c8d5/autoload/clap/floating_win.vim#L289-L301

And then you don't have to call `clap#layout#calc()` always, use `s:display_opts` instead, https://github.com/liuchengxu/vim-clap/blob/2ad4ce34a143b7734577388fba765d637f64c8d5/autoload/clap/floating_win.vim#L57

liuchengxu:
Try https://github.com/liuchengxu/vim-clap/tree/take-care-of-neovim-resized-event, should fix #454, the true solution needs you to be familiar with the code of clap.

autoload/clap/preview.vim
ray-x: This will make the float windows size `fixed` to preview_size instead of change to a smaller size if the search result is e.g. at 1st line.

autoload/clap/preview.vim
ray-x: One issue I saw is preview.show take buffer as arguments. The plugin should load buffer according to the number of lines need to be displayed. in this case is the number of lines recalculated based on windows size. 
A extreme example, if user set the preview_size to 1,000,000. The code should trim the size before performa file buffer load. 
Also, if user set the number to an incorrect value, clap#preview#size_of should return the number that was used by clap instead of the value use set.

liuchengxu: > A extreme example, if user set the preview_size to 1,000,000. The code should trim the size before performa file buffer load.

A plugin can't prevent you from doing evil things.

ray-x: The code should detect the incorrect setup in the first place.  clap#preview#size_of may not be the function, but it is also too to do that when plugin need to create the float window.

liuchengxu:
This patch doesn't fix the resizing issue. To reproduce, open neovim in a full tmux pane, 

image


and then split the pane horizonally.

This patch:

截屏2020-06-07 下午4 30 53

Branch https://github.com/liuchengxu/vim-clap/tree/take-care-of-neovim-resized-event:

截屏2020-06-07 下午4 31 27

You have to redo the whole layout to make it still usable.

And I want to explain the case of display window becoming small when you shrink the terminal size, that works as expected as I guess you are using a relative value in `g:clap_layout`. If you use a fixed size, the display window size won't change even your terminal size varies.

ray-x:
You are right, this PR does not handle vim window when clap opened.

liuchengxu:
Ok, I'll merge that branch if it also works for you. It's barely possible to handle all edge cases, what we can do is to cover most common cases :(.

ray-x:
Thanks! I am not sure which is the best way to handle that. In many GUI framework, the redraw/resize should trigger all component redraw/resize and each component will start document buffer reload on its own. But is kind of too much work for me. :-P

liuchengxu:
>In many GUI framework, the redraw/resize should trigger all component redraw/resize and each component will start document buffer reload on its own.

That's basically what that branch does, but the buffer reloading is needless, at least not handled by the plugin authors.

PR #465, opened 2020-06-10
Don't show Clap windows in `:Clap windows`

runiq:
Uses window IDs instead of buffers. Also fixes active window display.

### DIFF
--- a/autoload/clap/provider/windows.vim
+++ b/autoload/clap/provider/windows.vim
@@ -12,13 +12,10 @@ function! s:jump(t, w) abort
 endfunction
 
 function! s:get_clap_winids() abort
-  let clap_winids = []
-  for clap_win in ['display', 'input', 'spinner', 'preview']
-    let clap_win_var = 'g:clap.' . clap_win . '.winid'
-    if exists(clap_win_var)
-      call add(clap_winids, g:clap[clap_win].winid)
-    endif
-  endfor
+  let clap_winids = map(filter(
+          \ ['display', 'input', 'spinner', 'preview'],
+          \ 'exists("g:clap.".v:val.".winid")'),
+      \ 'g:clap[v:val].winid')
   if exists('g:__clap_indicator_bufnr')
     call extend(clap_winids, win_findbuf(g:__clap_indicator_bufnr))
   endif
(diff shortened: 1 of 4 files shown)


autoload/clap/provider/windows.vim
```diff
@@ -11,24 +11,45 @@ function! s:jump(t, w) abort
...
-  let modified = getbufvar(a:buf, '&modified')
-  let name = bufname(a:buf)
+function! s:get_clap_winids() abort
+	let clap_winids = []
+	for clap_win in ['display', 'input', 'spinner', 'preview']
+		let clap_win_var = 'g:clap.' . clap_win . '.winid'
+		if exists(clap_win_var)
+			call add(clap_winids, g:clap[clap_win].winid)
+		endif
+	endfor
```
liuchengxu, line 21:
```suggestion
let clap_winids = map(filter(['display', 'input', 'spinner', 'preview'], 'exists("g:clap.".v:val.".winid")'), 'g:clap[v:val].winid')
```

autoload/clap/provider/windows.vim
```diff
@@ -11,24 +11,45 @@ function! s:jump(t, w) abort
...
 
 function! s:windows.source() abort
+  let clap_winids = s:get_clap_winids()
   let lines = []
   for t in range(1, tabpagenr('$'))
-    let buffers = tabpagebuflist(t)
-    for w in range(1, len(buffers))
+    for w in range(1, tabpagewinnr(t, '$'))
+      " Skip Clap windows
+      let winid = win_getid(w, t)
```
liuchengxu, line 44:
```suggestion
    let winid = win_getid(w, t)
```

And please use Space instead of Tab for the other changes.

autoload/clap/provider/windows.vim
runiq: Thanks, that's going to be useful—didn't know (N)vim could do map.

runiq:
> Overall looks good, but I'm guessing you tried NeoVim only?

True, sorry. I thought that, since I only changed a single provider, the differences between Vim and Neovim would have already been abstracted away.

liuchengxu:
>the differences between Vim and Neovim would have already been abstracted away.

Most of them :(

runiq:
> One more needs to be fixed:

Done.

liuchengxu:
Thanks!

PR #467, opened 2020-06-18
Add clap#run()

### LINKED ISSUE #433: A way to execute ad-hoc searches similar to fzf
I dislike how with vim clap you need to pre-define each type of clap popup.  You cannot very easily dynamically create popups.

With fzf.vim, if you want to create a popup dynamically, you can just call `fzf#run` like this:
```viml
function! s:choseValue(value)
    echom "Chose value " . a:value
endfunction

call fzf#run({
    \ 'source': ['a', 'b', 'c'],
    \ 'down': '40%',
    \ 'sink': function('s:choseValue')})
```
Whereas, with vim-clap, it seems you have to do something like this:
```viml
function! s:choseValue(value)
    echom "Chose value " . a:value
endfunction

let g:clap_provider_temp1 = {
      \ 'source': ['a', 'b', 'c'],
      \ 'sink': function('s:choseValue')
      \ }

Clap temp1
```
This works ok, because you can define a temporary provider like this and re-use it for any adhoc popups you want to trigger, but it would be much nicer if there was an API method for this instead of needing to define a global variable for the temporary provider ahead of time

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

liuchengxu:
Close #433

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ CHANGELOG
 
 ## [unreleased]
 
+### Added
+
+- Add `clap#run(provider)` which is similar to `fzf#run()`. The argument `provider` is a Dict like `g:clap_provider_foo` with an optional extra field specifying the provider id. It can used for adhoc running, don't use it with a `source` that probably has a fair mount of items as it's normally undeveloped in performance. [#433](https://github.com/liuchengxu/vim-clap/issues/433)
+
 ### Fixed
 
 - Fixed the win contexted `execute()` for `jumps` and `marks` provider when clap window is not yet visible.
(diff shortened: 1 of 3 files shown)


runiq:
Quick question: If I *do* predefine a provider with specific settings and reference it by `id` in `clap#run()`, will the `clap#run()` function use that provider's settings (i.e. custom `on_move()`/`on_typed()` impls, cache behavior, etc.)?

liuchengxu:
Always use the provider's settings. The default impl injection only happens when the vital pieces are missing: https://github.com/liuchengxu/vim-clap/blob/7eb4c264a95156e9b799ef49d82349a1cfeb5023/autoload/clap.vim#L95-L117

runiq:
Thanks, that's exactly the bit of code I was looking for! :)

PR #474, opened 2020-06-30
Fix history provider open action support

zeroqn:
History provider supports open action, but action is not used in sink func.


liuchengxu:
Thanks!

PR #477, opened 2020-07-07
Provider preview for git commits and bcommits

ray-x:
Allow bcommits and commits to preview the changes (git log and git show)

### DIFF
--- a/autoload/clap/provider/bcommits.vim
+++ b/autoload/clap/provider/bcommits.vim
@@ -1,17 +1,81 @@
-" Author: liuchengxu <xuliuchengxlc@gmail.com>
 " Description: List the buffer commits.
 
 let s:save_cpo = &cpoptions
 set cpoptions&vim
 
 let s:bcommits = {}
+let s:bcommits.syntax = 'clap_diff'
+
+let s:current = bufname(g:clap.start.bufnr)
+let s:begin = '^[^0-9]*[0-9]\{4}-[0-9]\{2}-[0-9]\{2}\s\+'
 
 function! s:bcommits.source() abort
-  call g:clap.abort('Not implemented yet')
+  let s:git_root = clap#path#get_git_root()
+  if empty(s:git_root)
+    call g:clap.abort('Not in git repository')
+    return
+  endif
+
+  let s:current = bufname(g:clap.start.bufnr)
+  let managed = 0
+  if !empty(s:current)
+    call system('git show '.s:current.' 2> '.(has('win32') ? 'nul' : '/dev/null'))
+    let managed = !v:shell_error
+  endif
+
+  if !managed
+    call clap#helper#echo_error('The current buffer is not in the working tree' . s:current)
+  endif
+  let s:source = 'git log ''--color=never'' ''--date=short'' ''--format=%cd %h%d %s (%an)'' ''--follow'' ''--'' '.s:current
+  return split(system(s:source), "\n")
+endfunction
+
+
+function! s:bcommits.on_move() abort
+  let cur_line = g:clap.display.getcurline()
+  let sha=matchstr(cur_line, s:begin.'\zs[a-f0-9]\+' )
+
+  let l:prev = g:FindPrev(sha)
+  let gitdiff = 'git diff --color=never ' . l:sha . ' ' . l:prev . ' -- ' . ' '.s:current
+  let info = split(system(l:gitdiff), '\n')
+  if len(info) > 60
+    let info = info[:60]
+  endif
+
+  call clap#preview#show_with_line_highlight(info, 'diff', 0)
+  call clap#preview#highlight_header()
 endfunction
 
 function! s:bcommits.sink(line) abort
-  call g:clap.abort('Not implemented yet')
+  let s:current = bufname(g:clap.start.bufnr)
+  let sha=matchstr(a:line, s:begin.'\zs[a-f0-9]\+' )
+  let prev = g:FindPrev(sha)
+
+  let gitdiff = '!git diff --color=never ' . ' ' . sha . ' ' . prev . ' -- ' . s:current
+  vertical botright new
+  setlocal buftype=nofile bufhidden=wipe noswapfile nomodeline
+
+  setlocal modifiable
+  silent execute 'read' escape(gitdiff, '%')
+  normal! gg"_dd
+  setfiletype diff
+  setlocal nomodifiable
+endfunction
+
+function! g:FindPrev(ver) abort
+  let shas=split(system('git log --format=format:%h'), "\n")
+  let idx=0
+  let prev='master'
+  for commit in shas
+    if commit == a:ver
+      if idx + 1 < len(shas)
+        let prev = shas[idx+1]
+      endif
+      return prev
+    endif
+    let idx = idx+1
+  endfor
+  return prev
 endfunction
 
 let g:clap#provider#bcommits# = s:bcommits
--- a/autoload/clap/provider/bcommits.vim
+++ b/autoload/clap/provider/bcommits.vim
@@ -7,10 +7,12 @@ set cpoptions&vim
 let s:bcommits = {}
 let s:bcommits.syntax = 'clap_diff'
 
-let s:current = bufname(g:clap.start.bufnr)
+let s:current = ''
 let s:begin = '^[^0-9]*[0-9]\{4}-[0-9]\{2}-[0-9]\{2}\s\+'
+let s:shas = []
 
 function! s:bcommits.source() abort
+  let s:shas = split(system('git log --format=format:%h'), "\n")
   let s:git_root = clap#path#get_git_root()
   if empty(s:git_root)
     call g:clap.abort('Not in git repository')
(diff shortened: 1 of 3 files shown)


autoload/clap/provider/bcommits.vim
```diff
@@ -5,13 +5,78 @@ let s:save_cpo = &cpoptions
...
 function! s:bcommits.source() abort
-  call g:clap.abort('Not implemented yet')
+  let s:git_root = clap#path#get_git_root()
+  if empty(s:git_root)
+    call g:clap.abort('Not in git repository')
+    return
+  endif
+
+  let s:current = bufname(g:clap.start.bufnr)
+  let managed = 0
```
liuchengxu, line 21: `managed` is actually useless.

```vim
  if !empty(s:current)
    call system('git show '.s:current.' 2> '.(has('win32') ? 'nul' : '/dev/null'))
    if v:shell_error
      return ['The current buffer is not in the working tree' . s:current]
    endif
  endif
```

autoload/clap/provider/bcommits.vim
```diff
@@ -5,13 +5,78 @@ let s:save_cpo = &cpoptions
...
+  if !empty(s:current)
+    call system('git show '.s:current.' 2> '.(has('win32') ? 'nul' : '/dev/null'))
```
liuchengxu, line 23: I doubt `2> nul` can work on Windows.

autoload/clap/provider/bcommits.vim
```diff
@@ -5,13 +5,78 @@ let s:save_cpo = &cpoptions
...
+
+  call clap#preview#show_with_line_highlight(info, 'diff', len(info)+1)
+  call clap#preview#highlight_header()
 endfunction
 
 function! s:bcommits.sink(line) abort
-  call g:clap.abort('Not implemented yet')
+  let s:current = bufname(g:clap.start.bufnr)
+  let sha=matchstr(a:line, s:begin.'\zs[a-f0-9]\+' )
+  let prev = g:FindPrev(sha)
```
liuchengxu, line 53:
```suggestion
  let prev = s:find_prev(sha)
```

autoload/clap/provider/bcommits.vim
```diff
@@ -5,13 +5,78 @@ let s:save_cpo = &cpoptions
...
+  setlocal buftype=nofile bufhidden=wipe noswapfile nomodeline
+
+  setlocal modifiable
+  silent execute 'read' escape(gitdiff, '%')
+  normal! gg"_dd
+  setfiletype diff
+  setlocal nomodifiable
+endfunction
+
+function! g:FindPrev(ver) abort
```
liuchengxu, line 66:
```suggestion
function! s:find_prev(ver) abort
```

liuchengxu:
Thanks for the contribution! My concern is that this patch uses too many `system()` calls which run synchronously, not sure if it will become very slow for the repo having a ton of commits.

ray-x:
bcommit only works for a specific file. Normally should be fast to load all diff info. Not sure about how it going to work for some binary/large file with lots of commits.   I used a 40kb file(a project xml file) with ~7000 commit for test, seems not that bad.

autoload/clap/provider/bcommits.vim
ray-x: Fixed in new version

autoload/clap/provider/bcommits.vim
ray-x: Should be able to redirect. But I can not verify :-(

autoload/clap/provider/bcommits.vim
ray-x: Removed in new push

autoload/clap/provider/bcommits.vim
```diff
@@ -18,13 +18,7 @@ function! s:bcommits.source() abort
...
   let s:current = bufname(g:clap.start.bufnr)
-  let managed = 0
-  if !empty(s:current)
-    call system('git show '.s:current.' 2> '.(has('win32') ? 'nul' : '/dev/null'))
-    let managed = !v:shell_error
-  endif
-
-  if !managed
+  if empty(s:current)
     call clap#helper#echo_error('The current buffer is not in the working tree' . s:current)
```
liuchengxu, line 22: Return a List so that this error message can be shown in the display window.

```suggestion
    return ['The current buffer is not in the working tree' . s:current]
```

autoload/clap/provider/bcommits.vim
```diff
@@ -18,13 +18,7 @@ function! s:bcommits.source() abort
...
   endif
   let s:source = 'git log ''--color=never'' ''--date=short'' ''--format=%cd %h%d %s (%an)'' ''--follow'' ''--'' '.s:current
```
liuchengxu, line 24: Can the following `s:source` work in `system()`? That eems more readable.

```suggestion
  let s:source = 'git log --color=never --date=short --format="%cd %h%d %s (%an)" --follow --'.s:current
```
Or:

```suggestion
  let s:source = "git log '--color=never' '--date=short' '--format=%cd %h%d %s (%an)' '--follow' '--' ".s:current
```

autoload/clap/provider/bcommits.vim
```diff
@@ -36,7 +30,7 @@ function! s:bcommits.on_move() abort
   let cur_line = g:clap.display.getcurline()
   let sha=matchstr(cur_line, s:begin.'\zs[a-f0-9]\+' )
```
liuchengxu, line 31:
```suggestion
  let sha = matchstr(cur_line, s:begin.'\zs[a-f0-9]\+' )
```

autoload/clap/provider/bcommits.vim
```diff
@@ -5,13 +5,72 @@ let s:save_cpo = &cpoptions
...
+
+  setlocal modifiable
+  silent execute 'read' escape(gitdiff, '%')
+  normal! gg"_dd
+  setfiletype diff
+  setlocal nomodifiable
+endfunction
+
+function! s:find_prev(ver) abort
+  let shas=split(system('git log --format=format:%h'), "\n")
```
liuchengxu, line 61:
```suggestion
  let shas = split(system('git log --format=format:%h'), "\n")
```

And other several similar nits in this function :(.

autoload/clap/provider/bcommits.vim
```diff
@@ -5,13 +5,72 @@ let s:save_cpo = &cpoptions
 set cpoptions&vim
 
 let s:bcommits = {}
+let s:bcommits.syntax = 'clap_diff'
+
+let s:current = bufname(g:clap.start.bufnr)
```
liuchengxu, line 10: This line seems useless.

autoload/clap/provider/bcommits.vim
ray-x: This approach works 
`let s:source = "git log '--color=never' '--date=short' '--format=%cd %h%d %s (%an)' '--follow' '--' ".s:current`

autoload/clap/provider/bcommits.vim
ray-x: Fixed in new commits

autoload/clap/provider/bcommits.vim
ray-x: Yes, not really useful...

autoload/clap/provider/bcommits.vim
liuchengxu: I don't have a Windows machine either. I'm unwilling to merge a patch that possibly can't work on Windows. Need someone to verify that.

ray-x: Tried on a win10 VM, did not install vim, only tried git.  2> nul works fine

```
c:\>git show dumpfile
fatal: not a git repository (or any of the parent directories): .git

c:\>git show dumpfile 2> nul

c:\>
```

liuchengxu:
`commits` syntax is missing, fix this small issue then we can merge this.

```diff
diff --git a/autoload/clap/provider/commits.vim b/autoload/clap/provider/commits.vim
index 8b7fd66..ad0146d 100644
--- a/autoload/clap/provider/commits.vim
+++ b/autoload/clap/provider/commits.vim
@@ -5,7 +5,7 @@ let s:save_cpo = &cpoptions
 set cpoptions&vim
 
 let s:commits = {}
-let s:commits.syntax = 'clap_diff'
+
 function! s:commits.source() abort
   let s:git_root = clap#path#get_git_root()
   if empty(s:git_root)
@@ -66,7 +66,7 @@ function! s:commits.sink(line) abort
   setlocal nomodifiable
 endfunction
 
-let s:commits.syntax = 'clap_commits'
+let s:commits.syntax = 'clap_diff'
 
 let g:clap#provider#commits# = s:commits
```

ray-x:
commits.syntax will use 'clap_diff'. Code updated.

autoload/clap/provider/bcommits.vim
```diff
@@ -5,13 +5,78 @@ let s:save_cpo = &cpoptions
...
+  endif
+
+  let s:current = bufname(g:clap.start.bufnr)
+  if empty(s:current)
+    return ['The current buffer is not in the working tree' . s:current]
+  else
+    call system('git show '.s:current.' 2> '.(has('win32') ? 'nul' : '/dev/null'))
+  endif
+  let s:source = "git log '--color=never' '--date=short' '--format=%cd %h%d %s (%an)' '--follow' '--' ".s:current
+  return split(system(s:source), "\n")
```
liuchengxu, line 29: Actually you can directly return the system command, so that some native optimization of vim-clap can be used.

```suggestion
  return s:source
```

Same to the `source` of `commits` provider, the old code is not optimal.

```diff
+  return source
-  return split(system(source), "\n")
```

ray-x: Not sure if I missed anything here. After changed to 
```vim
let s:source = "git log '--color=never' '--date=short' '--format=%cd %h%d %s (%an)' '--follow' '--' ".s:current
return s:source
```
got an error message

```
Error detected while processing function <SNR>268_on_event[11]..<SNR>268_on_complete_maple:                         
line   22:                                                                                                          
E716: Key not present in Dictionary: lines, decoded.total)  
```
Are there any dependency?

liuchengxu: Are you using the latest vim-clap? Can't reproduce. Should work even without the Rust binary, so I would say there is no dependency here.

ray-x: Yes. I remove the folder and pull the vim-clap again. It works fine now.

autoload/clap/provider/commits.vim
```diff
@@ -31,14 +30,42 @@ function! s:commits.source() abort
   else
     let source .= ' --graph'
   endif
-  return split(system(source), "\n")
+  return source
+endfunction
+
+function! s:commits.sink(line) abort
+endfunction
+
```
liuchengxu, line 38:
```suggestion
```

liuchengxu:
Merged, thanks! I might impl the async version of them in Rust when I get some time.

PR #494, opened 2020-08-07
Fix `:Clap install-binary` does not work correctly on Windows

Bakudankun:
On Windows:
* Vim's `term_start()` executes a command directly (not on a shell), so you need to add `cmd.exe /c` to the command if you use `&&` etc.
* Still using nightly Rust.
* `cargo build` runs on wrong directory when building Python dynamic module.
* The name of the built dynamic module is wrong.

This is a PR to fix them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ CHANGELOG
 
 ## [unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


liuchengxu:
Thank you for fixing these issues!

PR #498, opened 2020-08-11
Document ClapFuzzyMatches highlighting group

oncomouse:
This is a minor issue, but I lost an hour yesterday trying to figure out how to highlight match results in fuzzy mode. This PR documents the highlight group for fuzzy matches.

### DIFF
--- a/doc/clap.txt
+++ b/doc/clap.txt
@@ -547,6 +547,11 @@ ClapMatches[1-8]                                                 *ClapMatches[1-
   `ClapMatches[1-8]` will be used for the matched substrings in the substring mode.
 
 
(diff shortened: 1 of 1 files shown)


liuchengxu:
Have you tried `:h g:clap_fuzzy_match_hl_groups`, you can use any number of highlight groups for fuzzy matched items, does not have to be `[1-12]`.

oncomouse:
It worked. I missed that, thanks!

I think I was confused by the matches highlighting is not being added as a key to the palette dictionary that is used for a lot of other highlighting stuff. Finding the right global variable makes total sense for how to set this, but it was a bit un-intuitive.

PR #505, opened 2020-08-19
Fix docs of g:clap_search_box_border_style

zhyu:
It seems `g:clap_search_box_border_style` will be used as a key of `g:clap_search_box_border_symbols`, which is a `Dict`. `g:clap_search_box_border_style` itself should be a `String`.


liuchengxu:
Thanks!

PR #534, opened 2020-10-01
Resolve file locking while updating binaries

### LINKED ISSUE #529: :Clap install-binary! don't work on Windows
If maple is already available (once downloaded) it will be started by Clap immediatly (at any startup routine, haven't examined). So maple.exe (for windows) is already running as process, therefore the downloaded (new version) can not been saved on the location (because the OS has a write lock on the running file),

Furthermore the plugin says the download was successful and the user can not examine the error log bacuse the split window has been closed so fast, you had no chance to read the message.

Solution:
If updating binary, temporary stop maple job,

Grueslayer:
Closes #529

### DIFF
--- a/autoload/clap/installer.vim
+++ b/autoload/clap/installer.vim
@@ -17,6 +17,9 @@ function! s:run_term(cmd, cwd, success_info) abort
       execute 'silent! bd! '.bufnr
       call clap#helper#echo_info(a:success_info)
     endif
+    if clap#maple#is_available()
+      call clap#job#daemon#start(function('clap#client#handle'))
+    endif
   endfunction
 
   if has('nvim')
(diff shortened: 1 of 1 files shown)


autoload/clap/installer.vim
```diff
@@ -106,8 +109,13 @@ endfunction
...
+    if executable(s:prebuilt_maple_binary)
+      "Since v0.7
+      let version_line = system(s:prebuilt_maple_binary.' version')
+      let s:current_version = str2nr(matchstr(version_line, '0.1.\zs\d\+'))
+    else
+      let s:current_version = -1
+    endif
   endif
   " Since v0.14 maple itself is able to download the latest release binary.
   if executable(s:prebuilt_maple_binary) && s:current_version >= 14
```
liuchengxu, line 121:
```suggestion
  if s:current_version >= 14
```

It's ensured `s:prebuilt_maple_binary` is executable when we are here.

liuchengxu:
Merged, thanks!

PR #541, opened 2020-10-14
[WIP] Smart-case

romgrk:
Hey,

This PR is a request for feedback regarding the implementation of smart-case filtering. As a reminder, in vim the `'smartcase'` option makes searching with a lower-case input case-insensitive, but whenever the input contains an uppercase letter the search becomes case-sensitive. This improves results a lot when there is a known upper-case letter in the filter input.

This is also listed as a [possible improvement](https://github.com/jhawthorn/fzy/blob/master/ALGORITHM.md#case-sensitivity) to the original fzy algorithm.

Without smart-case, the desired target in placed in 12th position by the scoring function:
Screenshot from 2020-10-14 19-07-01

With smart-case, the target in first position without unwanted targets:
Screenshot from 2020-10-14 19-07-19

This PR is a proof of concept, it only implements the filtering in the python scorer and is not optional. If you want to accept this PR I'll finish the implementation, here are more questions/comments:
 - IIUC, the filtering is done in `pythonx/clap/scorer.py`, `autoload/clap/filter/sync/viml.vim` and `autoload/clap/filter/async/dyn.vim`. Are there other places where the filtering mechanism should be updated?
 - I don't think this should be made optional, because case-insensitivity can be achieved by simply not using upper-case letters which is more costly in terms of typing, so no user should be doing it.

### DIFF
--- a/crates/matcher/extracted_fzy/src/lib.rs
+++ b/crates/matcher/extracted_fzy/src/lib.rs
@@ -10,9 +10,20 @@ use crate::scoring_utils::*;
 pub type MatchWithPositions = (Score, Vec<usize>);
 
 pub fn match_and_score_with_positions(needle: &str, haystack: &str) -> Option<MatchWithPositions> {
-    match matches(needle, haystack) {
+    let is_case_sensitive = match needle.chars().find(|c| c.is_uppercase()) {
+        Some(_) => true,
+        None => false,
+    };
+    let haystack_opt = if is_case_sensitive {
+        None
+    } else {
+        Some(haystack.to_lowercase())
+    };
+    let haystack_str = haystack_opt.as_deref().unwrap_or(haystack);
+
+    match matches(needle, haystack_str) {
         Some(needle_length) => {
-            let (score, positions) = score_with_positions(needle, needle_length, haystack);
+            let (score, positions) = score_with_positions(needle, needle_length, haystack_str);
             Some((score, positions))
         }
         None => None,
@@ -157,14 +168,10 @@ fn calculate_score(
     (D, M)
 }
 
-/// Compares two characters case-insensitively
+/// Compares two characters
 #[inline(always)]
 fn eq(a: char, b: char) -> bool {
-    match a {
-        _ if a == b => true,
-        _ if a.is_ascii() || b.is_ascii() => a.eq_ignore_ascii_case(&b),
-        _ => a.to_lowercase().eq(b.to_lowercase()),
-    }
+    a == b
 }
 
 fn compute_bonus(haystack: &str, haystack_length: usize) -> Vec<Score> {
--- a/crates/matcher/extracted_fzy/src/lib.rs
+++ b/crates/matcher/extracted_fzy/src/lib.rs
@@ -10,20 +10,15 @@ use crate::scoring_utils::*;
 pub type MatchWithPositions = (Score, Vec<usize>);
 
 pub fn match_and_score_with_positions(needle: &str, haystack: &str) -> Option<MatchWithPositions> {
-    let is_case_sensitive = match needle.chars().find(|c| c.is_uppercase()) {
-        Some(_) => true,
-        None => false,
-    };
-    let haystack_opt = if is_case_sensitive {
-        None
+    let haystack = if needle.chars().any(|c| c.is_uppercase()) {
+        haystack.into()
     } else {
-        Some(haystack.to_lowercase())
+        haystack.to_lowercase()
     };
-    let haystack = haystack_opt.as_deref().unwrap_or(haystack);
 
-    match matches(needle, haystack) {
+    match matches(needle, &haystack) {
         Some(needle_length) => {
-            let (score, positions) = score_with_positions(needle, needle_length, haystack);
+            let (score, positions) = score_with_positions(needle, needle_length, &haystack);
             Some((score, positions))
         }
         None => None,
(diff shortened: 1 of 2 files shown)


liuchengxu:
Look good to me, I agree making it the default behaviour.

I believe `viml` already supports the smart case, https://github.com/liuchengxu/vim-clap/blob/fc0022f69509c84f7d5901051f569cec682fc456/autoload/clap/filter/sync/viml.vim#L38

We should also see if the Rust version of fzy does the smart case. If you think the Python version of fzy also satisfies you, we can merge this now and I can also take a look at the Rust version later.
https://github.com/liuchengxu/vim-clap/blob/fc0022f69509c84f7d5901051f569cec682fc456/crates/matcher/src/algo.rs#L26

romgrk:
No I wanna do the rust version, finally an excuse to play with rust :D I'll send an update in a moment.

romgrk:
It's done. I've implemented it where you pointed me (`vim-clap/crates/matcher/src/algo.rs`), but there is something that's bothering me: doing it there means that `match_and_score_with_positions` checks *for each line* if needle contains uppercase characters. 

Ideally, the calling site should check if it contains uppercase characters and pass a `is_case_sensitive: bool` argument down the stack. Unfortunately, I seem to understand that multiple algorithms can be used, which causes indirection regarding the scoring function, which complicates adding a parameter because then we need to wrap the other algorithms (Skim and Substring). What do you think about that?

liuchengxu:
>Ideally, the calling site should check if it contains uppercase characters and pass a is_case_sensitive: bool argument down the stack.

Yeah, we need some abstraction for the current code so that we can make the behaviour consistent between these algorithms, but I recently don't have enough time to do this job, I think we could just improve fzy for now, it's ok for me.

crates/matcher/extracted_fzy/src/lib.rs
```diff
@@ -10,6 +10,17 @@ use crate::scoring_utils::*;
...
+        Some(_) => true,
+        None => false,
+    };
+    let haystack_opt = if is_case_sensitive {
+        None
+    } else {
+        Some(haystack.to_lowercase())
+    };
+    let haystack = haystack_opt.as_deref().unwrap_or(haystack);
+
```
liuchengxu, line 23: Can be simplified as:

```diff
diff --git a/crates/matcher/extracted_fzy/src/lib.rs b/crates/matcher/extracted_fzy/src/lib.rs
index d69f5da..03292bc 100644
--- a/crates/matcher/extracted_fzy/src/lib.rs
+++ b/crates/matcher/extracted_fzy/src/lib.rs
@@ -10,20 +10,15 @@ use crate::scoring_utils::*;
 pub type MatchWithPositions = (Score, Vec<usize>);

 pub fn match_and_score_with_positions(needle: &str, haystack: &str) -> Option<MatchWithPositions> {
-    let is_case_sensitive = match needle.chars().find(|c| c.is_uppercase()) {
-        Some(_) => true,
-        None => false,
-    };
-    let haystack_opt = if is_case_sensitive {
-        None
+    let haystack = if needle.chars().find(|c| c.is_uppercase()).is_some() {
+        haystack.into()
     } else {
-        Some(haystack.to_lowercase())
+        haystack.to_lowercase()
     };
-    let haystack = haystack_opt.as_deref().unwrap_or(haystack);

-    match matches(needle, haystack) {
+    match matches(needle, &haystack) {
         Some(needle_length) => {
-            let (score, positions) = score_with_positions(needle, needle_length, haystack);
+            let (score, positions) = score_with_positions(needle, needle_length, &haystack);
             Some((score, positions))
         }
         None => None,
```

liuchengxu: Can you add a Rust test for the smart case?

liuchengxu: Can you add a Rust test for the smart case?

romgrk:
Oh sure :) I've done the modification and added the test.

liuchengxu:
Merged, thank you!

crates/matcher/extracted_fzy/src/lib.rs
romgrk: @liuchengxu ok sorry to come back to this and sorry to bother but I have a question that just came to me :sweat_smile:

The way I did it the more complex way (with an `Option`) initially was because from what I read, `String` is allocated on the heap. I was wondering, isn't `haystack.into()` allocating things that would otherwise not be allocating by the way I initially implemented it? Or is it just taking the `&str` without allocating anything on the heap? I still have some trouble following how ownership works and where values are cleared. Thanks.

liuchengxu: Honstly, I'm also not sure about that 🙈 . After `hayhack.into()`, the ownership was moved, but I'm not sure if the compiler has some optimization tricks to not allocate. The true Rust Expert @ImmemorConsultrixContrarie might have a better answer.

ImmemorConsultrixContrarie: This could be simplified a little with a dark magic, but overall `Option` was a good shot.

```rust
let uppercase;
let haystack = if needle.chars().any(|c| c.is_uppercase()) {       
    uppercase = haystack.to_lowercase();
    &uppercase
} else {
    haystack
};
```

romgrk: Great, I had no idea you could create an undefined binding, that's neat :D Thanks to both for the precisions.

PR #546, opened 2020-10-17
Add background shadow

### LINKED ISSUE #545: Feature: background shadow when clap is active
Hi,

Would you accept a PR for a background-shadow when Clap is active?
I already have a proof-of-concept branch ready, visible below.

Screenshot from 2020-10-16 20-33-37

With preview:
Screenshot from 2020-10-16 20-46-44

romgrk:
Closes #545

Screenshot:

Screenshot from 2020-10-16 20-46-44


liuchengxu:
Hmm, so it's a neovim-only feature? Can we have a variable to control this in case some people are not fan of it.

autoload/clap/floating_win.vim
```diff
@@ -11,6 +11,9 @@ let g:clap#floating_win#display = {}
 let g:clap#floating_win#spinner = {}
 let g:clap#floating_win#preview = {}
 
+let s:shadow_bufnr = nvim_create_buf(v:false, v:true)
+let g:__clap_shadow_bufnr = s:shadow_bufnr
```
liuchengxu, line 15: `g: __clap_shadow_bufnr` is unused.

romgrk:
Forgot about vim :upside_down_face: 

 - There is no transparency in vim float windows right? :/ I'll add neovim guards in that case
 - Sounds good for the option

romgrk:
Done

liuchengxu:
Vim popup has a `popup-mask` feature, see `:h popup-mask`, not sure if it helps.

autoload/clap/themes.vim
```diff
@@ -162,6 +162,8 @@ function! s:reverse_PopupCursor() abort
 endfunction
 
 function! s:init_theme() abort
+  hi ClapShadow guibg=#000000
```
liuchengxu, line 165:
```suggestion
+  hi ClapDefaultShadow guibg=#000000
+  hi default link ClapShadow ClapDefaultShadow
```

So that user can easily override the highlight in their vimrc.

romgrk:
IIUC, `popup-mask` doesn't implement see-through transparency, it just hides those parts altogether :/

Screenshot from 2020-10-17 00-13-08

autoload/clap/themes.vim
```diff
@@ -162,6 +162,9 @@ function! s:reverse_PopupCursor() abort
 endfunction
 
 function! s:init_theme() abort
++  hi ClapDefaultShadow guibg=#000000
++  hi default link ClapShadow ClapDefaultShadow
```
liuchengxu, line 166:
```suggestion
   hi ClapDefaultShadow guibg=#000000
   hi default link ClapShadow ClapDefaultShadow
```

liuchengxu:
Hmm, seems there is no `winblend` alternative option for vim popup, so it can only have the background color, not the transparency like neovim's floating window. Anyway popup support does not have to be included in this PR, we could just merge this and wait for someone self sending another PR, but we should mention that it's a neovim-only option for now in the doc.

romgrk:
Fixed the typo and documented the shadow as neovim only. I've also added an option for the opacity because someone is going to want that option.

liuchengxu:
@romgrk Do you mind also updating the CHANGELOG.md?

romgrk:
Sure, it's updated

liuchengxu:
Merged, thank you!

PR #554, opened 2020-10-18
Add kPageUp/kPageDown bindings

hasufell:


liuchengxu:
What's the reason for including them as the default key mappings? I even can't find which key is `<kPageUp>` in my keybord. You can just add them in your vimrc if you like:

```vim
autocmd Filetype clap_input inoremap <silent> <buffer> <kPageDown> <C-R>=clap#navigation#scroll('down')<CR>
...
```

PR #557, opened 2020-10-20
Implement tagfiles provider

### LINKED ISSUE #446: Use pregenerated tag files
Ctags doesn't work well out of the box with Rust/Haskell/possibly many other languages. There are tools to generate these tag files (I have a plugin myself to handle the tools automatically on write), and Vim/Neovim is able to read them with the `tag` command. However, clap doesn't seem to know to use them instead.

Add a provider for searching preexisting tag files. For example, that provider should search the project root/current directory for tag files.

romgrk:
Closes #446

I chose to implement a separate provider (`tagfiles`) instead of improving `proj_tags` because the behavior & logic is too different and it's going to end up messy to try to keep both behaviors in a single provider.

I had some trouble understanding the logic of how vimscript & rust talk to each other so there are some hacks in this PR, I'll clean up before it's merged. It's also my first project with rust so the compiler and myself have been fighting all night. If you have any advice or see anything wrong with the rust code don't hesitate to point it out.

Tasks:
 - [x] Make the query only match in the tag name, not the whole line
 - [x] Ensure paths are shortened & truncated correctly
 - [x] Cleanup vim-rust communication
 - [x] Handle jumping to tag correctly, currently there might be conflicts if different tags have the same name
 - [x] Decide if we want to display tag's kind (either as text or as icon)
 - [x] Consider paths relative to tag file
 - [x] Shorten rendered paths if necessary
 - [x] Test parser correctness
 - [x] (?) FZY matches exact strings incorrectly
 - [x] Remove unrelated commits (refactoring)
 - [ ] Options: filter by language (aka file extension)

demo


romgrk:
I've also been doing some gentle refactoring on the vimscript side, and I may need to refactor a bit the rust side to decouple line searching and line rendering.

hasufell:
When I start searching in a 30mb tags file, after entering the first key, CPU load jumps to 100% and stays there for quite a while. The input becomes unresponsive and the chars show up one after one. The delay is several seconds.

Screenshot_2020-10-20_13-27-54

romgrk:
Got it, we probably need to be more lazy in the data exchange, or implement paging. I'll build some ginourmous tag files and check how to improve the performance before merging.

hasufell:
I tried the latest commit, it's much better!

romgrk:
Yes, the async workflow was already implemented I just didn't put it in place correctly because I don't know my way around the codebase yet :angel: 

This PR is mostly implemented, I still need to finish the handling of paths relative to their tagfile but if you use a tagfile at `./tags` it should be working fine. Please test it and report any bugs that you find.

I've implemented many hacks due to the rigidity of the vim-rust communication. First one is that instead of making the filtering algorithm take an `extracter` function to get the slice of text it needs to filter like I had envisionned, I simply made the `TAGS_RE` regex handle `tags`, `proj_tags` and `tagfiles` :/
The second one is that I need to send back to vim different informations: the rendered line, the file of the tag, and the pattern of the tag (to locate it within the file). However, the current protocol is limited to sending back a JSON object like `{ "lines": [...}, "indices": [...] }`. I didn't want to touch it because the architecture is very coupled and it needed more refactoring that I cared to do. So I just slapped a `:::FILENAME:::PATTERN` at the end of the rendered line and I `split(line, ':::')` it when I need the info back :laughing: I use the syntax to conceal the `:::...` part. I'm not proud but the first objective is to make it work.

hasufell:
With the last commit, I get

```
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
"~/.stack/indices/Hackage/packages/statistics/0.15.2.0/statistics-0.15.2.0/Statistics/Resampling/Bootstrap.hs" 119L, 4131B
vim-clap: target_dir:/home/jule/git/iohk/cardano-wallet-2, Run:function('280', {'source_type': 4, 'is_sync': function('299'), 'get_source': function('298'), 'init_default_impl': function('302'), 'can_async': function('301'), 'init_display_win': function('303'), 'support_multi_select': function('290'), '_apply_sink': function('280'), 'id': 'tagfiles', 'filter': function('289'), 'sink_star': function('283'), 'sink': function('282'), 'support_open_action': function('291'), '_': function('278'), 'is_pure_async': function('300'), 'try_set_syntax': function('293'), 'on_typed': function('285'), 'source_async_or_default': function('295'), 'on_exit': function('287'), 'has_enable_rooter': function('281'), 'source_async': function('296'), 'jobstop': function('288'), 'abort': function('279'), 'is_rpc_type': function('292'), 'apply_query': function('294'), 'on_enter': function('284'), '_apply_source': function('297'), 'on_move': function('286')}), run_args:['!Double   […e/packages/statistics/0.15.2.0/statistics-0.15.2.0/Statistics/Resampling/Bootstrap.hs]:::/home/jule/.stack/indices/Hackage/packages/statistics/0.15.2.0/statistics-0.15.2.0/Statistics/Resampling/Bootstrap.hs:::/^data T = {-# UNPACK #-} !Double :< {-# UNPACK #-} !Double$/;"'], exception:Vim(catch):E654: missing delimiter after search pattern: 
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
vim-clap: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jule/.cache/dein/repos/github.com/liuchengxu/vim-clap/crates/pattern/src/lib.rs:28:46
vim-clap: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

romgrk:
I've added a commit to handle failure in`tag_name_only`, which is supposed to extract the tagName from the string to limit the searching to it. It isn't supposed to fail because the input of `tag_name_only` is controlled by us. I've simply made it return the whole string if it fails because it's hard to debug which of the entries (that you didn't see) caused this panic. So if duringa tag search you see *more than the tag name* being searched in (highlighted), please take a screenshot of that and report it so I can fix it :)

hasufell:
Screenshot_2020-10-21_20-17-48

romgrk:
Awesome, could you drop me a copy of that tag file [here](https://www.dropbox.com/request/C6idiRxOwhMyySP2QMlA)?

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -19,40 +23,61 @@ struct TagInfo {
...
         let taken_width = name.len() + 1;
         let path_len = self.path.len() + 2;
         let mut adjustment = 0;
-        let path = if taken_width > winwidth {
-            format!("[{}]", self.path)
+
+        let mut home_path = PathBuf::new();
+        let path = Path::new(&self.path);
+        let path = path.strip_prefix(cwd).unwrap_or(
+            // FIXME: is there a way to avoid cloning HOME?
```
romgrk, line 35: @ImmemorConsultrixContrarie if you have any hints here (or on the PR in general) I'd be happy to have them :)

crates/maple_cli/Cargo.toml
```diff
@@ -18,6 +18,8 @@ structopt = "0.3"
 bytecount = "0.6"
 itertools = "0.9"
 serde_json = "1.0"
+dirs = "3.0"
+lazy_static = "1.4.0"
```
ImmemorConsultrixContrarie, line 22: [`once_cell`](https://docs.rs/once_cell/1.4.1/once_cell) is better.

crates/maple_cli/src/cmd/cache.rs
```diff
@@ -72,13 +72,9 @@ impl CacheEntry {
...
     /// info is cached in the file name.
     pub fn new(cmd_args: &[&str], cmd_dir: Option<PathBuf>, total: usize) -> Result<PathBuf> {
-        let mut dir = clap_cache_dir();
-        dir.push(cmd_args.join("_"));
-        if let Some(mut cmd_dir) = cmd_dir {
-            dir.push(format!("{}", calculate_hash(&mut cmd_dir)));
-        } else {
-            dir.push("no_cmd_dir");
-        }
+        let cmd_dir = cmd_dir.unwrap_or(PathBuf::from("no_cmd_dir"));
```
ImmemorConsultrixContrarie, line 75: Are you sure, that `no_cmd_dir` would not exist? Probably you should just go with

```rust
if let Some(cmd_dir) = cmd_dir {
    /* go code brr */
}
```

crates/maple_cli/src/cmd/tagfiles.rs
ImmemorConsultrixContrarie: `HOME.as_deref().map()`

romgrk:
So I've finished implementing path handling, which means that paths in tagfiles are now considered relative to where the tagfile is, which means that jumps work as expected. Any number of tagfiles can be fed to maple.

I've also added a global `--cwd` option that is passed to maple. This is the current directory in vim. It is used so maple can shorten paths to make them relative to where the user is. If that fails, paths are made relative to the user home directory (with a ~). If that fails, they're rendered as absolute.

Screenshot from 2020-10-23 03-16-42
(test with tags coming from different files, with a mix of asbolute & relative paths; all is well)

@hasufell I've haven't received any news :/ If you missed my comment above I requested if you could send me that tagfile that fails so I can debug the issue you're experiencing.

@liuchengxu There is one detail still bothering me (well for this PR), I've seen there is thic concept of "forerunner" in place, and even though I get the general idea of what we might want to do with that, the way `tagfiles` is setup (used `proj_tags` as a template), the forerunner process doesn't seem to do work that is used by the other processes, well at least when it's a huuuuuge (500mb) tagfile. It just keeps running and doesn't seem useful. I've also seen many different return values (`g:__clap_forerunner_tempfile`, `g:__clap_forerunner_result`, and nothing, which starts another maple) and it seems a bit much. I was wondering how you had envisionned all this and why there are so many options. If you would be able to explain a bit how it's supposed to work and how it could be improved I would like it very much :)

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -0,0 +1,260 @@
...
+        let path_label = if taken_width > winwidth {
+            format!("[{}]", path)
+        } else {
+            let available_width = winwidth - taken_width;
+            if path_len > available_width && available_width > 3 {
+                let diff = path_len - available_width;
+                adjustment = 2;
+                format!("[…{}]", path.chars().skip(diff + 2).collect::<String>())
+            } else {
+                format!("[{}]", path)
```
ImmemorConsultrixContrarie, line 59:
```rust
let path = path.display();            

let path_label = if taken_width > winwidth {        
    format!("[{}]", path)        
} else {        
    let available_width = winwidth - taken_width;        
    if path_len > available_width && available_width > 3 {        
        let diff = path_len - available_width;        
        adjustment = 2;        
        let path = path.to_string();            
        let path = path[path.char_indices().nth(diff + 2).map(|x| x.0).unwrap_or(path.len())..];
        format!("[…{}]", path)                                                       
    } else {        
        format!("[{}]", path)        
    }       
}; 
```

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    pub fn parse(base: &PathBuf, input: &str) -> Result<Self, std::string::ParseError> {
+        let mut field = 0;
+        let mut index = 0;
+        let mut index_last = 0;
+
+        let mut name = String::from("");
+        let mut path = String::from("");
+        let mut pattern = String::from("");
+        let mut kind = String::from("");
```
ImmemorConsultrixContrarie, line 88: `String::new` does the same and is easier to understand (at least for me).

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -0,0 +1,260 @@
...
+                    }
+                    2 => pattern = String::from(&input[index_last..index]),
+                    3 => kind = String::from(&input[index_last..index]),
+                    _ => {}
+                }
+                field += 1;
+                index_last = index + c.len_utf8();
+            }
+            index += c.len_utf8();
+        }
```
ImmemorConsultrixContrarie, line 106: Just use `char_indices`.

ImmemorConsultrixContrarie: Also, you allocate a new string each time you call a `String::from`. It is better to call `name.clear(); name.push_str(…);`.

ImmemorConsultrixContrarie: For `path` it would be

```rust
use std::fmt::Write;
path.clear();
write!(&mut path, "{}", path_buf.display()).unwrap();
```

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -0,0 +1,260 @@
...
+        .into_iter()
+        .map(move |path| read_tag_file(path, &cwd, winwidth))
+        .flatten();
+
+    let stream = Box::new(std::iter::empty()) as Box<dyn Iterator<Item = String>>;
+    let stream = streams.fold(stream, |acc, f| {
+        Box::new(acc.chain(f)) as Box<dyn Iterator<Item = String>>
+    });
+
+    Ok(stream)
```
ImmemorConsultrixContrarie, line 160: Just return
`files.into_iter().filter_map(move |path| read_tag_file(path, &cwd, winwidth).ok()).flatten()`

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -0,0 +1,260 @@
...
+            }
+        })
+    }))
+}
+
+fn create_tags_cache(
+    cwd: &PathBuf,
+    winwidth: usize,
+    args: &[&str],
+    files: &Vec<[PathBuf; 2]>,
```
ImmemorConsultrixContrarie, line 187: `&Vec<T>` is useless; always use `&[T]` in such places.

hasufell:
> I've haven't received any news :/

Sorry, was swamped with other stuff. I just uploaded a file

liuchengxu:
The `forerunner` job is spawned for collecting the whole results eagerly if `source` of provider is actually a shell command like `fd --type f` that needs to be executed in the terminal.

The `forerunner` job returns:

- the list of total lines if it's less 100000. https://github.com/liuchengxu/vim-clap/blob/b1f4e1d400209c3c97898b9b4caf577f93943923/crates/maple_cli/src/cmd/exec.rs#L25
- if the number of results lines is larger than 100000, they will be wrote to a cache file and vim will receive the cached file name. Next time when user inputs some query, clap will directly do the filtering on the cached file instead of spawning `source` cmd again.

You are right, it's unnecessary to spawn the foreruner job for `tagfiles` provider as the whole results are already there, we just need to do the filtering on the tagfiles. It's more like `blines` provider that querys against the current buffer which is in fact a file. You can implement `init` property in `tagfiles` provider explicitly instead of using the default impl so that `forerunner` won't be started.

https://github.com/liuchengxu/vim-clap/blob/b1f4e1d400209c3c97898b9b4caf577f93943923/autoload/clap/provider/blines.vim#L45

One improvement for `forerunner` job I can think of is to stop it properly. Currently the whole output will be cached in the memory that could be problem in some corner cases that you runs a crzay command, in which case we should terminate the forerunner job in time on the Rust side.

https://github.com/liuchengxu/vim-clap/blob/b1f4e1d400209c3c97898b9b4caf577f93943923/crates/maple_cli/src/light_command.rs#L161-L172

@romgrk Not sure if I make myself clear. Let me know if you have any questions.

autoload/clap.vim
```diff
@@ -85,16 +83,6 @@ let g:clap_multi_selection_warning_silent = get(g:, 'clap_multi_selection_warnin
 
 let g:clap_popup_border = get(g:, 'clap_popup_border', 'rounded')
 
-function! clap#builtin_providers() abort
```
liuchengxu, line 88: Can we extract these `tagfiles` provider unrelated code into another PR?

crates/maple_cli/src/cmd/tagfiles.rs
romgrk: Got it. For this particular case, the `.clear()` isn't necessary right?

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -0,0 +1,269 @@
...
+            "{text:<text_width$}{path_label}:::{path}:::{pattern}",
+            text = name,
+            text_width = text_width,
+            path_label = path_label,
+            path = self.path,
+            pattern = self.pattern,
+        )
+    }
+
+    pub fn parse(base: &PathBuf, input: &str) -> Result<Self, std::string::ParseError> {
```
liuchengxu, line 83: Since we are going to implement a parser to parse tags line, we might also want to take a look at the vimscript version of ctags parser, https://github.com/liuchengxu/vista.vim/blob/bcb50336188074c9cfc1aad4b463f4c0e698a68c/autoload/vista/parser/ctags.vim#L94 . It's theoratically fragile so I previously only support the ctags with JSON output in vim-clap instead of spending time doing this job, glad you eventually work on this :(.

autoload/clap.vim
romgrk: Will do

crates/maple_cli/src/cmd/tagfiles.rs
romgrk: Yes :)

romgrk:
Ok, so there are still a few things that need to be tested but basically the initial implementation is complete. The problem is I'm not satisfied with how things are. :(

Currently, how it works is as follow: when the user inputs a query, read the whole tag files, score each entry and return those that match. Whenever the query is updated, scratch the previously running process and start anew. It seems a bit brutal. Maybe I haven't understood some things correctly, but is this how the other similar providers are running?

I've been trying out the `tagfiles` provider in repos like the neovim source, with 24,000 entries, and it takes a few seconds before anything shows up. I would have expected things to be more performant. And the thing is I've tested the original C fzy on the command line, and when I search for something in that same 24,000 entries tagfile the search is, as far as my brain can tell, instant. This is what I expect from a fuzzy finder, not the current behavior. I will be investigating further how to improve because it's obvious that something is wrong with the current implementation.

fzy

@hasufell I couldn't find the entries listed in your screenshot in the tagfile you sent me :/ Could you check to include some tags that reproduce the problem illustrated? (eg I've `cat tags | grep PACKAGE_VERSION` but didn't find anything)

liuchengxu:
>I've been trying out the tagfiles provider in repos like the neovim source, with 24,000 entries, and it takes a few seconds before anything shows up

It's probably because you are running some loop in vimscript, basically all the heavy jobs should be done in Rust, and vim only  receives the final results without doing any processing but displaying them. I can take a look later, `blines` provider is a better template for `tagfiles`.

romgrk:
Yes, there is probably something looping. I'll look into it in the next days.

ImmemorConsultrixContrarie:
It could be a natural [disk latency](https://gist.github.com/jboner/2841832), not some vimscript loop. Just saying.

crates/maple_cli/src/cmd/tagfiles.rs
```diff
@@ -145,27 +156,25 @@ pub struct TagFiles {
 fn read_tag_files<'a>(
     cwd: &'a PathBuf,
     winwidth: usize,
-    files: &'a Vec<[PathBuf; 2]>,
+    files: &'a [[PathBuf; 2]],
 ) -> Result<impl Iterator<Item = String> + 'a> {
-    let streams = files
+    Ok(files
         .into_iter()
```
ImmemorConsultrixContrarie, line 162: `into_iter` on a shared reference is the same as `iter`; the latter gives less surprising results and is better here.

romgrk:
Screenshot from 2020-10-27 04-32-47

Lol well that wasn't hard to find.

crates/maple_cli/src/cmd/cache.rs
romgrk: Not sure what you mean here. Also I've removed all code dealing with caching for tagfiles as it's not useful in this case. Should I revert that?

romgrk:
Ok ok, so I think I'm mostly done now.

As indicated by my last comment, performance issues were due to me trying to use (complex) regexes as string parsers which is always a bad idea. Hopefully I'll remeber now. However, to fix that issue I had to introduce something a bit hacky; the part of maple that fetches items can only provide a list of String to the filterer, without additional metadata. To know where to parse, I've introduced a separator token (`::::`) that marks starts/ends of locations in the strink. The `tag_name_only` parser now uses at first. As both `tagfiles` and `proj_tag` go through the same branches, I've updated `proj_tags` to use it as well. I don't like it, but the other solution is to try to refactor maple so that fetching & filtering can exchange metadata and I have been fighting with rustc way too much right now.

The problem with `tag_name_only` previously is that tag names, even if in theory they shouldn't contain spaces, contains spaces in practice (even universal-ctags does it), quite unsuprinsigly. The only character that I'm sure can't be in the tag name is the tab character. Tag names can be surprising. I even worry that Haskell operators could be `::::`. I'm sure it's possible but let's ignore that possibility for now.

I have also implement a more robust tag line parser. Most of the fields are pretty easy to parse, but the tagaddress is the most problematic one because it can contain escaped characters, so we need to do real parsing instead of just finding start/end indexes with tabs. I've added many tests coming from different tag tools: universal-ctags, scss-ctags, and hasktags (@hasufell I couldn't install codex and if you see my comment above your tagfile didn't seem to contain the invalid lines...)

There is one detail that I've left out, maple's tagfiles subcommand doesn't handle the `--languages` subcommand, which would be nice to handle. Usually I know for which language I'm looking a tag for and that can narrow the search. OTOH I'm low on energy, and maple is performant again due to the fix above, so I haven't felt annoyed with speed even in 25k+ tags directories. I'd maybe keep it asa future improvements.

Comments welcome.

liuchengxu:
>the part of maple that fetches items can only provide a list of String to the filterer, without additional metadata. 

That can be a good direction.

>introduced a separator token (::::) that marks starts/ends of locations in the strink.

Hmm, not a fan of `::::`. But I'm also not sure if the way of vista.vim is robust enough https://github.com/liuchengxu/vista.vim/blob/bcb50336188074c9cfc1aad4b463f4c0e698a68c/autoload/vista/parser/ctags.vim#L91-L92

>" {tagname}\<Tab>{tagfile}\<Tab>{tagaddress}[;"\<Tab>{tagfield}..]
>" {tagname}\<Tab>{tagfile}\<Tab>{tagaddress};"\<Tab>{kind}<Tab>{scope}

Basically only `tagaddress` is not easy to parse, maybe we can simply parse out `tagname` and `tagfile` and then deal with the rest fields. @masatake Any tips for parsing these fields from the tagfile?

romgrk:
> That can be a good direction.

Yes that's how I would envision things for the future but it will require a substantial refactoring. During this PR I at some point started such refactoring but I ended up lost in a confusion of types (because it's my first rust project). Not sure how long that may take.

> Hmm, not a fan of ::::

Me neither. But I couldn't see any intermediary steps between that and the refactor. I'm open to propositions.

> " {tagname}<Tab>{tagfile}<Tab>{tagaddress}[;"<Tab>{tagfield}..]
> " {tagname}<Tab>{tagfile}<Tab>{tagaddress};"<Tab>{kind}{scope}
> 
> Basically only tagaddress is not easy to parse, maybe we can simply parse out tagname and tagfile and then deal with the rest fields.

For your 2 examples, as long as `tagaddress` is valid, my parse will work (I'll test it to be sure). There is quite a bit of variation in what characters come after `tagadress`, but actually all we need to do is to first parse the valid address, then skip any non-tab characters that come at the end. A valid address means either:
 - a series of digits
 - a search pattern starting and ending with a slash `/`; `\` escapes slashes and itself

Anything else than those 2 possibilities is an invalid addres an is rejected. IMHO the tag parser is acceptable as it is, it can always be adjusted as bugs are reported.

masatake:
I'm bad in English. So I read only the part of this pull request. I will answer happily if you summarize your issue and open an issue for summarized one at universal-ctags/ctags repo. In addition I don't know vimscript at all. I'm an Emacs user.

```
" {tagname}<Tab>{tagfile}<Tab>{tagaddress}[;"<Tab>{tagfield}..]
" {tagname}<Tab>{tagfile}<Tab>{tagaddress};"<Tab>{kind}{scope}
```

> Basically only tagaddress is not easy to parse, maybe we can simply parse out tagname and tagfile and then deal with the rest fields. @masatake Any tips for parsing these fields from the tagfile?

My first question is "do you need the value of {tagaddress}"?
If you don't need it, you should not generate it.
With -n option you can suppress it:
```
$ cat foo.c
cat foo.c
int
main(void)
{
	return 0;
}
$ ./ctags -o - foo.c
./ctags -o - foo.c
main	foo.c	/^main(void)$/;"	f	typeref:typename:int
$ ./ctags -o - -n foo.c
./ctags -o - -n foo.c
main	foo.c	2;"	f	typeref:typename:int
```
The regex pattern for the tagaddress becomes simple: `[0-9]+`.

If you need it and if you need precise result, you must parse ` {tagaddress}` carefully.
See https://docs.ctags.io/en/latest/man/tags.5.html.

If you can accept false-positive-alike result, you can use following pattern:
```
([^\t]+)\t([^\t]+)(.*)\$/(;"\t(.*)$)?
```
Here
\1 -> tagname, \2->tagfile, \3$/"->tagaddress, \4->rest fields.

hasufell:
> @hasufell I couldn't install codex and if you see my comment above your tagfile didn't seem to contain the invalid lines...

@romgrk There are no invalid lines. My point was that it searches both the filepath and the identifier. But I haven't tested the latest commit.

hasufell:
I tried the latest commit and it doesn't really work for me. I don't even know where it searches, but it can't find regular identifiers. When I use `:ts <ident>` then I get correct results though, so the tags file is correct.

liuchengxu:
>I will answer happily if you summarize your issue and open an issue for summarized one at universal-ctags/ctags repo.

Sure, just opened https://github.com/universal-ctags/ctags/issues/2683 :)

autoload/clap/api.vim
```diff
@@ -165,6 +169,10 @@ function! s:init_display() abort
     endfunction
 
     function! display.set_lines(lines) abort
+      if has_key(g:clap.provider._(), 'preprocess')
```
liuchengxu, line 172: Can we not add `preprocess` for `::::`, just process these stuffs on the Rust side?

romgrk: Ok, I'll review the PR to either do the refactor, or preprocess on the Rust side

romgrk:
> If you can accept false-positive-alike result, you can use following pattern:
> 
> ([^\t]+)\t([^\t]+)(.*)\$/(;"\t(.*)$)?
>
> Here
> \1 -> tagname, \2->tagfile, \3$/"->tagaddress, \4->rest fields.

We need the tagaddress and the regex won't work for us as we don't support just universal-ctags. The current parse accepts the following. I've reviewed the man page for tags, and IMHO our parser is complete enough. Here are the different address that are accepted by the parser:
 - pattern such as `/function main()/` plus any non-tab chars (such as `;"` or `[;"`)
 - pattern such as `/^\tfunction main() \/\/ comment$/` plus any non-tab chars
 - line numbers such as `83` plus any non-tab chars
One thing in the man page not accepted by the parser would be multiple search pattern such as:
 - `/struct foo/;/int x/;`
I think accepting multiple search patterns add unnecessary complexity, and I haven't seen anyone generating such addresses in practice. It's not that much more work but IMHO not necessary. Let me know what you think.

@hasufell Could you add the run it with the following path? It will place the search command (including tagfiles paths) in your clipboard. You could also run the command directly in your terminal to check the output. I'd be happy if you could report any issue that you find.

<details>
<summary>patch.txt</summary>
<pre>
diff --git a/autoload/clap/provider/tagfiles.vim b/autoload/clap/provider/tagfiles.vim
index 9e1344c..741d634 100644
--- a/autoload/clap/provider/tagfiles.vim
+++ b/autoload/clap/provider/tagfiles.vim
@@ -53,6 +53,7 @@ function! s:update_query()
   let args += ['tagfiles', g:clap.input.get()]
   let args += map(tagfiles(), {_, f -> '--files=' . f})
   let cmd = call(function('clap#maple#build_cmd'), args)
+  let @+ = cmd
   call clap#filter#async#dyn#start_directly(cmd)
 endfunc
 
</pre>
</details>

romgrk:
Sorry about the huge commit it's not self-consistent otherwise. Here are the changes:
 - Sources now return a iterator over `SourceItem { display: String, filter: Option<String> }` instead of `String`. The optional `filter` field is filled with what needs to be matched in, and `display` represents the rendered line. For now only tag providers use `filter`
 - There were cyclic dependencies problems so I extracted commonly re-used types out of `printer` and into their own crate, `types`. Which I noticed after that it conflicted with `stdio_server`'s `types` which has been renamed `server_types`.
 - I've done it as simply as I could because I still have some trouble with Rust borrow-checker. There are a few `XXX` comments in places where I think cloning should be avoidable but am not sure how to do so. (ping @ImmemorConsultrixContrarie ?)
 - The refactor feels very good. I was able to eliminate the `tag_name_only` function all its friends could probably be eliminated ruthlessly.

masatake:
> line numbers such as 83 plus any non-tab chars

There is a case violating the above assumption (in u-ctags output).

```
$ cat /tmp/bar.c
int main (void) {	return 0; } /* -> \treturn */
$ ./ctags --excmd=combine -o - /tmp/bar.c 
main	/tmp/bar.c	0;/^int main (void) {	return 0; } \/* -> \\treturn *\/$/;"	f	typeref:typename:int
```

liuchengxu:
>The refactor feels very good. I was able to eliminate the tag_name_only function all its friends could probably be eliminated ruthlessly.

@romgrk I like the direction 👍 , it's definitely making vim-clap extensible, but can we extract these refactorings into some smaller PRs to make the review easier? For example, We can make the `types` refactoring a seperate PR, and then we can first experiment the tagfile parser in the current `proj_tags`, which can be another PR. After that, we can focus on the new `tagfiles` provider integration.

romgrk:
> @romgrk I like the direction +1 , it's definitely making vim-clap extensible, but can we extract these refactorings into some smaller PRs to make the review easier? 

I'd be happy to but as far as I could see it was less optimal.

> For example, We can make the types refactoring a seperate PR, and then we can first experiment the tagfile parser in the current proj_tags, which can be another PR. After that, we can focus on the new tagfiles provider integration.

To experiment in both `tagfiles` and `proj_tags` we needed a way to match the tag_name without the `::::` separator. This was impossible other than the last commit refactoring. The previous implementation (regex matching) wouldn't have worked because tag names can contain whitespaces, digits and colons, all of which would have confused `tag_name_only`.

The types refactoring was necessary to the above refactoring because the types were defined in the `printer` crate but they were needed in all crates. This refactoring imho simple enough, it was a copy-paste of [these lines](https://github.com/liuchengxu/vim-clap/pull/557/commits/54255e51944f278131a56afbabd9e90fa4a3c84b#diff-43597a3c260647a67283feb2ed8d6e5b365b2c2483dcada97fc6e32cb46d3402L10-L26) into their own crate. The `server_types` refactor is unfortunate though, but happened mid-way through after renaming lots of `printer::` to `types::`. I can reverse the `server_types` refactor if you insist and have a good name for my `types` crates. This name needs to reflect the love I have for it.

Separating this PR in different PRs also has disadvantages, for examples `proj_tags` and `tagfiles` would need different extractor, which means we would need to keep `tag_name_only` and also create more applier functions in `matcher`'s `line_splitter.rs`, and also I think creating another LineSplitter symbol throughout the codebase.

Well I hope my arguments for why it's not desirable to split this PR have convinced you, otherwise let me know I will review for how it could be split. Also if your hesitations are partly due to the robustness of the parse I have no problem improving it, adding tests, and researching the different tag formats. I will address the u-ctags issue mentioned above.

crates/filter/src/source.rs
```diff
@@ -7,34 +8,40 @@ use subprocess::Exec;
...
+impl From<Vec<String>> for Source<std::vec::IntoIter<SourceItem>> {
     fn from(source_list: Vec<String>) -> Self {
-        Self::List(source_list.into_iter())
+        // XXX: Not sure about the .into_iter().map().collect().into_iter(), pretty sure it's doing
+        // something wasteful.
+        Self::List(source_list
+            .into_iter()
+            .map(|line| SourceItem { display: line, filter: None })
+            .collect::<Vec<_>>()
+            .into_iter())
```
ImmemorConsultrixContrarie, line 27: I don't like this impl at all. Probably `Source` should have a function like this:

```rust
fn string_into_source_item(s: String) -> SourceItem;

pub fn from_iter<I>(iter: I) -> Source<std::iter::Map<I::IntoIter, string_into_source_item>>
where I: IntoIterator<Item = String>
{
    Self::List(iter.into_iter().map(string_into_source_item))
}
```

ImmemorConsultrixContrarie: Still not the best thing to do, but I'm too lazy to fix it properly.

crates/filter/src/source.rs
```diff
@@ -61,18 +68,22 @@ impl<I: Iterator<Item = String>> Source<I> {
...
-                    matcher(&line).map(|(score, indices)| (line.into(), score, indices))
+                    matcher(&line).map(|(score, indices)| (SourceItem::new(line.into()), score, indices))
                 })
                 .collect::<Vec<_>>(),
             Self::List(list) => list
-                .filter_map(|line| matcher(&line).map(|(score, indices)| (line, score, indices)))
+                .filter_map(|item| {
+                    // XXX: How do I avoid cloning here?
+                    let line = item.filter.clone().unwrap_or(item.display.clone());
+                    matcher(&line).map(|(score, indices)| (item, score, indices))
```
ImmemorConsultrixContrarie, line 85: You probably missed that one in the `unwrap_or` function docs, so here it is: whenever you do something heavy in `_or` part, use `unwrap_or_else(|| /* heavy part */ )`. Right now you clone `display` part on *every* item, and then, if `filter` part is `Some`, it is dropped.

```rust
matcher(item.filter.as_deref().unwrap_or(&item.display)).map(|(score, indices)| (item, score, indices))
```

crates/matcher/src/line_splitter.rs
```diff
@@ -34,67 +35,90 @@ impl From<String> for LineSplitter {
     }
 }
 
+// XXX: avoid cloning everywhere below. Cloning is bad.
+
 #[inline]
-pub(super) fn apply_on_grep_line_skim(line: &str, query: &str) -> MatcherResult {
-    strip_grep_filepath(line).and_then(|(truncated_line, offset)| {
+pub(super) fn apply_on_grep_line_skim(item: SourceItem, query: &str) -> Option<FilterResult> {
+    strip_grep_filepath(&item.display.clone()).and_then(|(truncated_line, offset)| {
```
ImmemorConsultrixContrarie, line 42: You don't need a `clone` here at all, just `&item.display`

crates/matcher/src/line_splitter.rs
```diff
@@ -34,67 +35,90 @@ impl From<String> for LineSplitter {
...
-            .map(|(score, indices)| (score, indices.into_iter().map(|x| x + offset).collect()))
+            .map(|(score, indices)| (item, score, indices.into_iter().map(|x| x + offset).collect()))
     })
 }
 
 #[inline]
-pub(super) fn apply_on_tag_line_skim(line: &str, query: &str) -> MatcherResult {
-    tag_name_only(line).and_then(|tag_name| fuzzy_indices_skim(tag_name, query))
+pub(super) fn apply_on_tag_line_skim(item: SourceItem, query: &str) -> Option<FilterResult> {
+    fuzzy_indices_skim(&item.filter.clone().unwrap(), query)
```
ImmemorConsultrixContrarie, line 90: `item.filter.as_deref().unwrap()`
The same goes for two things down there.

crates/matcher/src/line_splitter.rs
```diff
@@ -34,67 +35,90 @@ impl From<String> for LineSplitter {
     }
 }
 
+// XXX: avoid cloning everywhere below. Cloning is bad.
```
ImmemorConsultrixContrarie, line 38: You can try to delete all `clone()` for `item.display`. I don't see why types should not match here without `clone()`. With `item.filter` it is just `item.filter.as_deref()` and then you do whatever you want with `Option<&str>`.

crates/filter/src/dynamic.rs
```diff
@@ -330,12 +335,16 @@ macro_rules! source_iter_file {
...
 macro_rules! source_iter_list {
     ( $scorer:ident, $list:ident ) => {
-        $list.filter_map(|line| $scorer(&line).map(|(score, indices)| (line, score, indices)))
+        $list.filter_map(|item| {
+            // XXX: How do I avoid cloning here?
+            // let line = item.filter.clone().unwrap_or(item.display.clone());
+            $scorer(item)
+        })
     };
 }
```
ImmemorConsultrixContrarie, line 344:
```rust
macro_rules! source_iter_list {
    ( $scorer:expr, $list:expr ) => {
        match ($list, $scorer) {
            (list, scorer) => list.filter_map(|item| {
                let line = item.filter.unwrap_or(item.display);
                scorer(&line).map(|(score, indices)| (line, score, indices))
            })
        }
    };
}
```

If you need to preserve the item instead of line, then it would be

```rust
macro_rules! source_iter_list {
    ( $scorer:expr, $list:expr ) => {
        match ($list, $scorer) {
            (list, scorer) => list.filter_map(|item| {
                scorer(item.filter.as_deref().unwrap_or(&item.display)).map(|(score, indices)| (item, score, indices))
            })
        }
    };
}
```

crates/filter/src/source.rs
romgrk: Yes I had guessed but I needed your input on the other cloning thing so I did the simplest thing I could. Seems you often solve my cloning issues with `.as_deref()`, I'll read on that.

crates/matcher/src/line_splitter.rs
romgrk: The borrow-check was complaining about partially moved `item`. I'll check what I can remove without making him angry.

crates/filter/src/dynamic.rs
romgrk: Good to know! It's a left-over of an initial transformation but I did run into that problem during that initial transformation.

crates/filter/src/source.rs
ImmemorConsultrixContrarie: There's no magic, `as_deref` is just a function with `match &x { Some(x) => Some(x.deref()), None => None }` but writing `as_deref` is just simpler than writing `match`. Same goes for `as_ref()`, `as_mut` and [other `Option` methods](https://doc.rust-lang.org/std/option/enum.Option.html) — they all could be implemented with `match`, no magic.

liuchengxu:
@romgrk @ImmemorConsultrixContrarie I have made a PR #583 about the fetching&filtering metadata direction, I'm also not the fan of `SourceItem` way. When we want to display or match something special, use a special matcher then. Feel free to leave any comments.

romgrk:
Already saw :) Left a comment directly on that PR, we can continue the discussion there for the matchers architecture.

liuchengxu:
Not sure if you still have some interest in this PR, it's probably a pain to resolve the conflicts now -_-, you might want to start over on top of the latest master. I'm happy to help push it forward if you still have a plan to work on it. @romgrk

romgrk:
TBH I've been using [coc-clap](https://github.com/vn-ki/coc-clap) for symbol-jumping lately, and I am low on time these days :/

I'll close it for now as it would indeed be too much work to resolve conflicts, but we can always use some of it later if the tagfiles is implemented.

liuchengxu:
@romgrk Totally understand it. From my point of view, `proj_tags` is actually already fast enough.

hasufell:
That's a pretty bad sign for the contribution workflow to let a PR like this go stale. I'm still stuck on this branch.

romgrk:
Unfortunate but open source development is limited by the time and interest of contributors :| You're welcome to take over the PR.

hasufell:
> Unfortunate but open source development is limited by the time and interest of contributors :| You're welcome to take over the PR.

I would, but I'm having health issues that limit the time I can code. Plugins that improve ergonomics are key for me.

What I meant with contribution workflow is this: On any of my projects, I'd never introduce major changes to `master` while a PR as big as this is still open. I'd either ping the author of the PR until it's done or take over the remaining work.

Avoiding major merge conflicts is up to the maintainer of the repository, imo, not to contributors. It's hard to have contributors in open source at all. Difficulties with getting things merged just scares them off.

PR #562, opened 2020-10-21
Update clap.txt

jackpotrykus:
The current documentation to change keybindings for insert-mode scroll in a clap window did not work on my system.

This works -- but it doesn't remove the functionality of <C-j> (i.e. i can use either <C-n> or <C-j>, and likewise <C-p> or <C-k>)


romgrk:
Cool, thanks :)

PR #564, opened 2020-10-23
Implement animations

romgrk:
Ok so I know this sounds crazy but it's not as bad as you think it is :laughing: I made two GIFs with different applications but both feel jankier than it feels in real. The blend (opacity) animation does feel nice, but I must admit the display-height animation feel a bit janky no matter what because it's limited to increasing by a line at each step :/

demo

clap-animate

(I only half-expect you to take this one seriously, I don't expect people to be enthusiastic with animations, feel free to decline ^^)

### DIFF
--- a/autoload/clap/animate.vim
+++ b/autoload/clap/animate.vim
@@ -1,0 +1,76 @@
+
+let g:clap_animation_frequency = get(g:, 'clap_animation_frequency', 50)
+
+function! clap#animate#start(duration, initial, final, type, Fn)
+  let duration = a:duration
+  let initial  = a:initial
+  let final    = a:final
+  let type     = a:type
+
+  let ticks = (duration / g:clap_animation_frequency) + 10
+
+  let state = {}
+  let state.running = v:true
+  let state.Fn = a:Fn
+  let state.type = type
+  let state.step = (final - initial) / ticks
+  let state.duration = duration
+  let state.current = initial
+  let state.initial = initial
+  let state.final = final
+  let state.start = reltime()
+  let state.timer = timer_start(
+    \ g:clap_animation_frequency,
+    \ {timer -> s:animate_tick(timer, state)},
+    \ { 'repeat': ticks })
+
+  call state.Fn(state.timer, state.current)
+  return state
+endfunc
+
+function! s:animate_tick(timer, state)
+  let state = a:state
+
+  " Alternative to finding current value:
+  "
+  "   let state.current += state.step
+  "   call state.Fn(a:timer, current)
+  "
+  " The reason why I go the long way (below) is because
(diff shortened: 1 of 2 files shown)


liuchengxu:
Nice job! Looks super cool, but we need to ensure it won't introduce any performance regression before doing the merge.

autoload/clap/animate.vim
```diff
@@ -0,0 +1,76 @@
+
+let g:clap_animation_frequency = get(g:, 'clap_animation_frequency', 50)
+
+function! clap#animate#start(duration, initial, final, type, Fn)
+  let duration = a:duration
+  let initial  = a:initial
+  let final    = a:final
+  let type     = a:type
```
liuchengxu, line 8: This does not look very helpful, just using `a:*` is good.

autoload/clap/animate.vim
```diff
@@ -0,0 +1,76 @@
...
+  call timer_stop(a:state.timer)
+endfunc
+
+function! clap#animate#lerp(ratio, initial, final, ...)
+  let type = a:0 > 0 ? a:000[0] : v:t_number
+
+  let range = a:final - a:initial
+  let delta = type == v:t_number ?
+    \ float2nr(a:ratio * range) :
+    \          a:ratio * range
```
liuchengxu, line 71: This line is not too long, no need to wrap it?

```suggestion
  let delta = type == v:t_number ? float2nr(a:ratio * range) : a:ratio * range
```

autoload/clap/animate.vim
romgrk: Yes I was playing with the parameters format (between a dict or like that), I'll remove those.

romgrk:
> Nice job! Looks super cool, but we need to ensure it won't introduce any performance regression before doing the merge.

Yes! This is really just a POC. The opacity animation works nice, but the height animation needs to be handled correctly. The height can be changed mid-animation by the `display.shrink*` methods, that's why I commented them for this. I'll need to model it that in a way that is clearer and ensure the animation adapts to any new requested height. I'll probably need to hide the `nvim_win_set_config(s:display_windi` calls behind a `s:display_set_config` that will check the animation status before applying the height. I'll complete the tags PR first though.

HaleTom:
This looks very sweet!  Would you be up for looking at it again with fresh 2022 eyes?

romgrk:
@HaleTom I don't have as much time for OSS these days :| Also I'm not interested enough in continuing this PR, I have other stuff to play with right now (integrating kitty graphics into neovim), but whoever wants to continue the work is welcome to take it.

haskelcurry:
Hey just wanted to say it's awesome, thanks @romgrk

romgrk:
Closing, stale.

PR #566, opened 2020-10-24
Fix file preview and Clap filer in vim8

ryo7000:
When `job_start()` was first called by vim-clap in vim8, preview and Clap filer were not working.
The reason is that the job id in vim8 starts at 0, but the condition of whether the job exists or not is wrong.

To check the first id of the job in vim8, I ran the following command to verify it.
```
vim -u NONE -U NONE
echo ch_info(job_getchannel(job_start('ls')))
```
The above command outputs the following

```
{'status': 'open', 'id': 0, 'in_io': 'pipe', 'err_mode': 'NL', 'in_status': 'open', 'out_io': 'pipe', 'err_io': 'pipe', 'err_status': 'open', 'out_timeout': 2000, 'out_status': 'open', 'out_mode':
 'NL', 'in_mode': 'NL', 'in_timeout': 2000, 'err_timeout': 2000}
```

So I changed the conditions for the existence of job in vim8.

### DIFF
--- a/autoload/clap/dispatcher.vim
+++ b/autoload/clap/dispatcher.vim
@@ -11,7 +11,7 @@ let s:job_id = -1
 let s:drop_cache = get(g:, 'clap_dispatcher_drop_cache', v:true)
 
 function! s:jobstop() abort
-  if s:job_id > 0
+  if s:job_exists()
     call clap#job#stop(s:job_id)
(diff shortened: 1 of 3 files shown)


romgrk:
LGTM. I've been able to reproduce the issue with vim, and the PR fixes it.

PR #570, opened 2020-10-26
Improve command provider

romgrk:
Quick improvements for the default `command` provider:
 - Style: Parse & render each command instead of feeding `execute('command')` directly into clap
 - Behavior: if a command has 0 arguments, it is executed immediately; otherwise, the user is put in command-mode with `:CommandName<space>` pre-filled so they can enter additional arguments
 - Style: commands with 0 arguments have their argument highlighted with `WarningMsg` to denote they will be executed immediately

@liuchengxu This is ready for merging as it is, I wanted to do more but the rest can wait.

Before:
Screenshot from 2020-10-26 17-04-48

After:
Screenshot from 2020-10-26 16-55-40

### DIFF
--- a/autoload/clap/provider/command.vim
+++ b/autoload/clap/provider/command.vim
@@ -8,14 +8,6 @@ set cpoptions&vim
 let s:command_header   = v:null
 let s:command_list = []
 
-function! s:right_pad(s, n)
-    return a:s . repeat(' ', a:n - len(a:s))
-endfunction
-
-function! s:left_pad(s, n)
-    return repeat(' ', a:n - len(a:s)) . a:s
-endfunction
-
 function! s:sink(selected) abort
   " :h command, note the characters in the first two columns
   let cmd = matchstr(a:selected, '\v^\s*\S+\s+\zs\S+')
@@ -50,7 +42,7 @@ let g:clap#provider#command# = s:command
 
 " Helpers
 
-function! s:parse_command (input)
+function! s:parse_command(input) abort
   let bang = stridx(a:input[0:3], '!') != -1
   let input = a:input[4:]
   let name = matchstr(input, '\v^\S+')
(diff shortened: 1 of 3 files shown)


romgrk:
No wait a few details to finish.

romgrk:
Ok, now I'm done. Below you can observe the result. I also wanted to discuss a bit rendering in general. I think it feels nice to have the start of the matched text align with the cursor start. I was thinking about introducing the concept of left-margin in the codebase. `margin` would be the width of text before the matched text in the display window. This margin can be used for many different purposes depending on each provider. For `command`, it's the arity of the command. But for others it could be:
 - `*tag*`: the kind (letter or icon) of the tag
 - `files`: the filetype icon
 - `blines`: the filepath & lnum
 - etc
This margin would have a minimum, let's say 3 spaces. This would anyway be dictated by the length of the provider name, `command_history` creates a bigger margin than `tags`. Or it could be adjusted/configured for each provider, eg `bline` or `grep` showing filepath+lnum might want a bigger margin. Anyway let me know what you think about this idea.

And the PR is ready for real now.

Screenshot from 2020-10-26 17-50-44

autoload/clap/provider/command.vim
```diff
@@ -4,21 +4,74 @@
 let s:save_cpo = &cpoptions
 set cpoptions&vim
 
-let s:command = {}
 
-function! s:command.sink(selected) abort
+let s:command_header   = v:null
+let s:command_list = []
+
+function! s:right_pad(s, n)
```
liuchengxu, line 11:
```suggestion
function! s:right_pad(s, n) abort
```

autoload/clap/provider/command.vim
```diff
@@ -4,21 +4,74 @@
...
+    return a:s . repeat(' ', a:n - len(a:s))
+endfunction
+
+function! s:left_pad(s, n)
```
liuchengxu, line 15:
```suggestion
function! s:left_pad(s, n) abort
```

autoload/clap/provider/command.vim
```diff
@@ -4,21 +4,74 @@
...
+let s:command = {}
 let s:command.syntax = 'clap_command'
+let s:command.source = function('s:source')
+let s:command.sink = function('s:sink')
 
 let g:clap#provider#command# = s:command
 
+" Helpers
+
+function! s:parse_command (input)
```
liuchengxu, line 53:
```suggestion
function! s:parse_command(input) abort
```

liuchengxu:
Looks good to me, just fix the lint and then we can merge this.

I like the `margin` idea, but we have to make it adaptive rather then being objective to the width of provider name.

romgrk:
All done.

Yes the margin would be adaptative. I'll open a PR with a draft once I have a better idea of what I want.

liuchengxu:
Merged, thank you! 🍻 

My quick idea is to introduce a new attribute `margin` for each provider, which should also be passed to the filter, e.g., 

```vim
let s:command.margin = 9
```

If no `margin` attribute, treat the margin as 0.

PR #577, opened 2020-11-01
fix: trigger autocmmands correctly when restoring window

romgrk:
clap-bug

This fixes a bug that I noticed in [barbar.nvim](https://github.com/romgrk/barbar.nvim). Barbar has a feature that a new buffer is opened next in order to the current buffer. The current clap behavior for restoring the current window is:
 - Close clap window, letting vim decide which window to focus back *triggering autocommands*
 - If the window isn't the right one, switch windows *without triggering autocmd* (`clap.start.goto_win()`)

This makes it that the last current buffer from any other plugin's perspective is the one to which vim focused back. This solves the issue.

### DIFF
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -182,12 +182,13 @@ endfunction
 
 " Sometimes we don't need to go back to the start window, hence clap#_exit() is extracted.
 function! clap#exit() abort
-  call clap#_exit()
+  noautocmd call clap#_exit()
 
   " NOTE: Need to go back to the start window
   if win_getid() != g:clap.start.winid
     call g:clap.start.goto_win()
   endif
+  doautocmd WinEnter
 endfunction
 
 function! clap#should_use_raw_cwd() abort
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -182,12 +182,8 @@ endfunction
 
 " Sometimes we don't need to go back to the start window, hence clap#_exit() is extracted.
 function! clap#exit() abort
+  noautocmd call g:clap.start.goto_win()
   noautocmd call clap#_exit()
-
-  " NOTE: Need to go back to the start window
-  if win_getid() != g:clap.start.winid
-    call g:clap.start.goto_win()
-  endif
   doautocmd WinEnter
 endfunction
 
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -155,7 +155,7 @@ function! clap#_exit() abort
   call clap#job#regular#forerunner#stop()
   call clap#maple#clean_up()
 
-  call g:clap.close_win()
+  noautocmd call g:clap.close_win()
 
   let g:clap.is_busy = 0
   let g:clap.display.cache = []
@@ -182,9 +182,8 @@ endfunction
 
 " Sometimes we don't need to go back to the start window, hence clap#_exit() is extracted.
 function! clap#exit() abort
-  noautocmd call g:clap.start.goto_win()
-  noautocmd call clap#_exit()
-  doautocmd WinEnter
+  call g:clap.start.goto_win()
+  call clap#_exit()
 endfunction
 
 function! clap#should_use_raw_cwd() abort
--- a/autoload/clap/api.vim
+++ b/autoload/clap/api.vim
@@ -8,7 +8,7 @@ let s:is_nvim = has('nvim')
 let s:cat_or_type = has('win32') ? 'type' : 'cat'
 
 function! s:_goto_win() dict abort
-  noautocmd call win_gotoid(self.winid)
+  call win_gotoid(self.winid)
 endfunction
 
 function! s:_getbufvar(varname) dict abort
--- a/autoload/clap/provider/blines.vim
+++ b/autoload/clap/provider/blines.vim
@@ -9,7 +9,6 @@ let s:blines = {}
 function! s:blines.sink(selected) abort
   let lnum = matchstr(a:selected, '^\s*\(\d\+\) ')
   let lnum = str2nr(trim(lnum))
-  call g:clap.start.goto_win()
   " Push the current position to the jumplist
   normal! m'
   silent call cursor(lnum, 1)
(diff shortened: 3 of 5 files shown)


autoload/clap.vim
```diff
@@ -182,12 +182,13 @@ endfunction
...
 " Sometimes we don't need to go back to the start window, hence clap#_exit() is extracted.
 function! clap#exit() abort
-  call clap#_exit()
+  noautocmd call clap#_exit()
 
   " NOTE: Need to go back to the start window
   if win_getid() != g:clap.start.winid
     call g:clap.start.goto_win()
   endif
+  doautocmd WinEnter
```
liuchengxu, line 191: I think this patch should be enough?

```diff
diff --git a/autoload/clap.vim b/autoload/clap.vim
index 70b0897..b4ae9ac 100644
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -187,6 +187,7 @@ function! clap#exit() abort
   " NOTE: Need to go back to the start window
   if win_getid() != g:clap.start.winid
     call g:clap.start.goto_win()
+    doautocmd WinEnter
   endif
 endfunction
```

romgrk: It does work yes, on the other hand if we let autocommands being triggered during `clap#_exit()` I feel that we're adding unnecessary work. Because there will be autocommands triggered for staying in a window that will not keep the focus. Those autocommands perform work, such as updating tabline and statusline plugin. It's small delays, but it adds to the perceived perfomance of clap because it increases the time before opening the new buffer.

I also have another proposition:
```vim
" Sometimes we don't need to go back to the start window, hence clap#_exit() is extracted.
function! clap#exit() abort
  noautocmd call g:clap.start.goto_win() " noautocmd is repeated in goto_win() but is here for readability
  noautocmd call clap#_exit()
  doautocmd WinEnter
endfunction
```

I think this might be even better because it is functionally equivalent, and it completely avoids focusing a window that isn't needed, we go directly back to the previous window, then cleanup, then trigger the autocommand.

liuchengxu: >Those autocommands perform work, such as updating tabline and statusline plugin.

That's my convern. I'm not 100% sure whethe it's appropriate to disable these autocmds, some plugins might be broken like your barbar.nvim, so I previously keep using `noautocmd` as much as less even I do care about the performance. I'm not saying I'm against it, it's definitely good for the performance. Let's try it for a while and see if something expected happens.

romgrk: And what was the rationale for marking `clap.start.goto_win()` as `noautocmd`, if you remember? And do you have some example use-cases of plugins where this would cause perturbations? I could try a few of them see that we're not causing unexpected behavior. IMHO the current implementation is good, I've been studying autocommands in the last few days due to this plugin and [that PR](https://github.com/neovim/neovim/pull/13117) and I'm confident in the current behavior.

One thing I've realized just now is that there are other exit points from clap such as `clap#selection#try_open`. Let's leave this open I'll test them in the next days.

liuchengxu: >And what was the rationale for marking clap.start.goto_win() as noautocmd, if you remember? 

Hmm, if I recall correctly, using `noautocmd` inside `g:clap.start.goto_win()` is simply to avoid writing `noautocmd` when calling `g:clap.start.goto_win()` :). So adding `doautocmd WinEnter` is indeed what I missed before.

>And do you have some example use-cases of plugins where this would cause perturbations?

https://github.com/liuchengxu/vim-clap/blob/9a89a257a713b93e309c1ab82d38085f0185bed9/autoload/clap.vim#L158

If someone has some `WinClosed` autocmd config in his vimrc, then I believe it won't be triggered.

romgrk:
I did some more testing with the different providers and I've explored the code and the flow of window exit. I've changed a bit the implementation, here is the flow now:
 - `clap.start.goto_win()` triggers autocommands
 - `clap#_exit()` does not trigger autocommands
 - we don't use `doautocmd WinEnter` anymore because `start.goto_win()` has done the job natively

The end goal here is, when the sink is called, both the sink and any plugins must consider that they are in the current buffer's window, just as if the user was typing `:edit ...` in the command line and coming back. Removing `noautcmd` from `start.goto_win` allows that to happen whenever the `start.goto_win() .... code ... clap#_exit()` pattern happens. The only place where is was beneficial (but not necessary) to add back `noautocmd` was in the `colors` provider, but that provider is already doing something funky to start with so he should expect to have more work.

I've also removed a few `start.goto_win()` from some providers' sink. When the sink is called, it can and should be assumed that the focus has been placed back in the original window.

This is ready for merging if you're ok with the implementation.

Note: there seems to be a bug with some providers with icons (tested grep here): multi-select + ctrl-t opens the filepath including the icon. I'll do some more testing with the others and file an issue

autoload/clap.vim
```diff
@@ -182,12 +182,8 @@ endfunction
 
 " Sometimes we don't need to go back to the start window, hence clap#_exit() is extracted.
 function! clap#exit() abort
+  call g:clap.start.goto_win()
```
liuchengxu, line 185:
```suggestion
  noautocmd call g:clap.start.goto_win()
```

How about using `noautocmd` here? Feels like we never leave the start window...

romgrk: No, that's precisely the moment when the autocommands will be triggered. It's because we do it without autocmd that other plugins can update their state correctly. We could add it here, but then we'd need to add `doautocmd WinEnter`. Doing `doautocmd WinEnter` is also only partially correct, because plugins (or the user config) might be listening to other autocmds (`BufEnter`). Letting autocommands be triggered natively is also more future-proof.

One alternative that would really feel to plugins as if the user never leaves the start window is to put `noautocmd` everywhere: from when clap opens, until when clap exits. This would also have the advantage of being more performant because no other scripts can run. However, this is a more involved modification and I don't feel ready to do it at this point.

liuchengxu: Ok, that makes sense.

liuchengxu (approved):
LGTM.

PR #583, opened 2020-11-07
Split out matchers

liuchengxu:
Introduce a trait `MatchItem` for various needs of getting the matching
text from the origin line.

### DIFF
--- a/crates/matcher/src/lib.rs
+++ b/crates/matcher/src/lib.rs
@@ -19,6 +19,7 @@
 
 mod algo;
 mod line_splitter;
+mod matchers;
 
 pub use algo::*;
 pub use line_splitter::*;
--- a/crates/matcher/src/line_splitter.rs
+++ b/crates/matcher/src/line_splitter.rs
@@ -1,8 +1,9 @@
 use crate::{
     fzy, skim::fuzzy_indices as fuzzy_indices_skim, substring::substr_indices, MatcherResult,
 };
-use pattern::{file_name_only, strip_grep_filepath, tag_name_only};
 use structopt::clap::arg_enum;
+
+use crate::matchers::{FileNameMatcher, GrepMatcher, MatchItem, TagNameMatcher};
 
 // Implement arg_enum so that we could control it from the command line.
 arg_enum! {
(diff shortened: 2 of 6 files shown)


romgrk:
I feel that it won't be efficient to implement it this way. For example, in the tagfiles thing the for-matching text (tag name) and the rendered line are computed before they are sent to the `apply_on_x` functions. With the changes here, we will run into the same problem that the `TagNameMatcher` needs to parse the rendered line. That's why it's necessary to pass as input more that just the rendered line, the input type of `apply_on_x` functions needs to change.

(Sorry in advance if you've already thought about that, just wanted to share this eventual problem I see)

liuchengxu:
>For example, in the tagfiles thing the for-matching text (tag name) and the rendered line are computed before they are sent to the apply_on_x functions.

Hmm, I'm not sure about this. Currently we always run `apply_on_x(origin_line, query)` on the origin line, not the rendered line, and outputs `(origin_line, match_score, matched_indices)`. Am I missing somthing?

romgrk:
Well if you check for example, `tags.rs`:

https://github.com/liuchengxu/vim-clap/blob/992c2d43318ba096eedac0592a0c2d6f90fed236/crates/maple_cli/src/cmd/tags.rs#L150-L157

`formatted_tags_stream`, which generates the stream of lines for `Source::List` and `filter::dyn_run`, provides a stream of rendered tag lines (line 90 at `tag.format()` is the line renderer for tags):

https://github.com/liuchengxu/vim-clap/blob/992c2d43318ba096eedac0592a0c2d6f90fed236/crates/maple_cli/src/cmd/tags.rs#L84-L97

Haven't looked for real at the other providers though, but for this one, as far as I understand, we're sending rendered lines down the pipeline.

liuchengxu:
Well, if you were saying `tag.format()`, that's what I meant the `origin_line` :). `tags` now only supports the JSON lines and does not need the extra parser, we have to build the _normal_ line based on the JSON and then feed them into the `filter::dyn_run`.

For the `tagfiles` provider in #557, we need to run our own parser to parse the raw content to build the _normal_ line, which can also be seen as the `origin_line`, but you're right if we can save some metadata then we don't have to parse again in the filter/display steps.

romgrk:
> if we can save some metadata then we don't have to parse again in the filter/display steps

Indeed. IMHO it would be also way cleaner, because each source would be responsible of saying "here is my display line", and "here is my text to match on". The current architecture tries to give the responsability of finding the to-be-matched area downstream, but it doesn't transmit the necessary metadata to do so.

Putting the responsability of creating the metadata (filter line, match area, and any other thing we might want in the future) also simplifies considerably all the `apply_on_xxx_algo(item, query)`, they simply become `apply_algo(itemWithMetadata, query)`, and below simply needs one case per algorithm.

https://github.com/liuchengxu/vim-clap/blob/992c2d43318ba096eedac0592a0c2d6f90fed236/crates/matcher/src/lib.rs#L36-L55

I tell you my friend, I really feel it's a cleaner way to do it :smile:

liuchengxu:
Absolutely agree, let's see if we could find a nice solution for this situtation, but really busy these days, I can only find some scattered time for the open source projects 🙈

romgrk:
> but really busy these days, I can only find some scattered time for the open source projects

Ya I get it. It's tiring to have a regular influx of issues. And people always expect you to solve their issues :/

> let's see if we could find a nice solution for this situation

Well the tagfiles PR does go into the direction I described ^^ But I feel that you disagree with certain aspects of it. My first question would be are you ok with the idea of source producing iterators over `SourceItem` instead of `String`? I don't really see another way to implement what I described. Otherwise, how would you structure it? Let me know what you think, and don't hurry if you don't have time, anyway the PR is doing what I made it for on my machine so I'm happy for now.

liuchengxu:
>would be are you ok with the idea of source producing iterators over SourceItem instead of String? 

I'm ok with the direction, but not sure the form of `SourceItem`. I wonder if using generic `SourceItem: From<String> + MatchItem` is good for this task.

```rust
/// Source is anything that can produce an iterator of String.
#[derive(Debug)]
pub enum Source<SourceItem: From<String> + MatchItem, I: Iterator<Item = SourceItem>> {
    Stdin,
    #[cfg(feature = "enable_dyn")]
    Exec(Exec),
    File(PathBuf),
    List(I),
}
```

Each of `Source` can produce a stream of `String` at the very beginning, and then each item is transformed to some specific `SourceItem` which has its own filter/display text other than the initial `String`. But I haven't tried this way :(.

romgrk:
> `SourceItem: From<String> + MatchItem`

You completely lost me on this one ^^ I'm not yet up to date on Rust types. What are the benefits brought from that form in terms of performance or extensibility, by opposition to using a simple struct? Meanwhile Ima do some googling on that `type + type` form...

liuchengxu:
>What are the benefits brought from that form in terms of performance or extensibility, by opposition to using a simple struct?

Using a simple struct does work, but it's not flexible enough in my opinion as we might want to use different structs for the various cases, using the generic can do this job as long as the struct implements the bounded traits, just want to explore whether it's doable, but I can also accept a simple struct if we don't have enough time to dig into this :).

romgrk:
> The generic seems not worth it compared to the much more complexity it adds after I simply tried

I trust you on that! Seems like a common problem in rust ^^ The good thing with the simple struct is that it's easily modifiable. It also fits all of our use-cases currently, I don't see what we could need other than the rendered line and the text to match.

liuchengxu:
Closed due to https://github.com/liuchengxu/vim-clap/pull/611

PR #599, opened 2020-12-06
Initial support for fzy lua

### LINKED ISSUE #565: Support fzy in lua
Now clap already supports `fzy` in pure Python, Python dynamic module in Rust and pure Rust, but none of them is native to neovim. And lua is generally much faster than Python, especically neovim uses luajit. There is already a fzy implementation in lua https://github.com/swarn/fzy-lua, [related telescope.nvim PR](https://github.com/nvim-lua/telescope.nvim/pull/184), I think it's great to also support it in clap. We merely need some glue code like https://github.com/liuchengxu/vim-clap/blob/master/autoload/clap/filter/sync/python.vim for this task.

liuchengxu:
Close #565

Thanks to https://github.com/swarn/fzy-lua, the basic fzy Lua support in clap is much easier. The next step might be adding a user-friendly way to also support your https://github.com/romgrk/fzy-lua-native @romgrk . Let me know if you have any ideas about that.

~I have tested this PR against the latest neovim, but not sure I'll spend more time making it work with the older neovim and vim.~

Now it works with both latest neovim and vim.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ target
 testdata.txt
 publish.sh
 crates/Cargo.lock
+pythonx/clap/fuzzymatch_rs.so
 
 !.editorconfig
 !.travis.yml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add `g:clap_popup_move_manager` so that Vim users can override the default mappings easily. #536
 - Allow user to always download the prebuilt binary. #531
 - Support smartcase fitlering for fzy algo and it's the default behavior. #541 @romgrk
+- Add initial support for fzy lua. #599
 
 ### Improved
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Add `g:clap_popup_move_manager` so that Vim users can override the default mappings easily. #536
 - Allow user to always download the prebuilt binary. #531
 - Support smartcase fitlering for fzy algo and it's the default behavior. #541 @romgrk
-- Add initial support for fzy lua. #599
+- Add initial support for fzy lua, neovim-nightly or vim compiled with lua is required. #599
 
 ### Improved
 
--- a/autoload/clap/debugging.vim
+++ b/autoload/clap/debugging.vim
@@ -59,7 +59,11 @@ function! clap#debugging#info() abort
   echohl Normal | echon has('python3')           | echohl NONE
 
   echohl Type   | echo 'has py dynamic module: '                | echohl NONE
-  echohl Normal | echon clap#filter#sync#python#has_dynamic_module() | echohl NONE
+  try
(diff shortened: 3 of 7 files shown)


romgrk:
I've added a single export point to `fzy-lua-native`, so you should be able to use it by simply requiring the file at `fzy-lua-native/lua/init.lua`. This file will automatically try to load the native version and fallback to the pure-lua one if that fails. So by loading it you can then use `fzy.score` or `fzy.positions`.

Eg
```lua
local fzy = dofile('.../path/to/fzy-lua-native/lua/init.lua')
if fzy.has_match(...) then
 ...
end
```
I use `dofile` instead of `require` because it's easier to load a file by absolute path that way.


Note that both the original and native implementation expose a `fzy.filter(needle, haystacks, case_sensitive)` function but it's format differs a bit from the `fzy.filter` that you have implemented here. If you want it can be modified for our usage, I think the folks at telescope are not using `fzy.filter` (@tjdevries can you confirm :+1: if we can modify that API? Dont wanna break it for you)

One more thing, I was playing with improvements and have added a `fzy.filter_many` in the native version, what it does is it loops over the haystacks in C rather than in lua. Haven't had time to make it the default but it probably should be because it brings a 100% improvement over the native-but-loop-in-lua version:
Screenshot from 2020-12-07 17-56-57

For including the package, I'd either copy-paste everything or include as a git submodule. I prefer copy-paste to avoid complexifying the install process of vim-clap.

Last thing, fzy-lua-native includes pre-compiled binaries for linux (x64 and arm), macOS (x64) and windows (x64), so it should be good for most users, but we might want to add a build step in `Clap install-binary` for this. All that is needed is to run `make` in the root of fzy-lua-native with gcc installed. I can probably add support for clang if it's there.

tjdevries:
@romgrk you can break however is best. 😁 I have your thing as submodule, so it's locked to a particular version. I can upgrade and fix differences as you go.

liuchengxu:
@romgrk I have added the Lua support for Vim too. I'll leave the fzy-lua-native integration in the feature PR :(. I prefer to use it as a submodule as multiple files including the various prebuilt `.so` have to be copied and can bloat the clap repo?

romgrk:
:+1: 

TBH fzy-lua-native isn't super useful for clap because there is already the rust backend for the heavy-lifting

liuchengxu:
I might be just too greedy :)

ttys3:
this is really awesome!

show current filter impl in lightline plugin:

```vim
function! ClapFilterImplStatusLine() abort
   let cur_impl = clap#filter#current_impl()
   return 'Clap: ' . get({
         \   'Lua': luaeval("require'nvim-web-devicons'.get_icon('', 'lua')"),
         \   'Python': luaeval("require'nvim-web-devicons'.get_icon('', 'py')"),
         \   'VimL': luaeval("require'nvim-web-devicons'.get_icon('', 'vim')"),
         \}, cur_impl)
endfunction
```

image

PR #611, opened 2021-01-10
Refactor each input line as a SourceItem

liuchengxu:
In this PR we use `SourceItem` to represent each input line which is useful when we need another form for displaying/matching the original line. The implementation is not perfect, just want to make some progress here :(. More details: #577

cc @romgrk @ImmemorConsultrixContrarie


crates/filter/src/dynamic.rs
```diff
@@ -128,13 +130,13 @@ fn try_notify_top_results(
...
-                let (text, _, idxs) = std::ops::Index::index(buffer, idx);
+                let (item, _, idxs) = std::ops::Index::index(buffer, idx);
                 let text = if let Some(painter) = icon_painter {
                     indices.push(idxs.iter().map(|x| x + ICON_LEN).collect::<Vec<_>>());
-                    painter.paint(&text)
+                    painter.paint(item.display_text())
                 } else {
                     indices.push(idxs.clone());
-                    text.clone()
+                    item.display_text().to_owned()
```
liuchengxu, line 139: `&str` -> `String`, the lucky thing is that the item size is limited.

crates/filter/src/source.rs
```diff
@@ -7,34 +7,28 @@ use subprocess::Exec;
...
     Exec(Exec),
     File(PathBuf),
     List(I),
 }
 
-impl From<Vec<String>> for Source<std::vec::IntoIter<String>> {
-    fn from(source_list: Vec<String>) -> Self {
-        Self::List(source_list.into_iter())
-    }
-}
```
liuchengxu, line 22: Haven't found a good alternative for this :(.

crates/filter/src/source.rs
```diff
@@ -61,18 +55,21 @@ impl<I: Iterator<Item = String>> Source<I> {
...
-                        matcher(&line).map(|(score, indices)| (line, score, indices))
+                        matcher(&line).map(|(score, indices)| (line.into(), score, indices))
                     })
                 })
                 .collect::<Vec<_>>(),
             Self::File(fpath) => std::fs::read_to_string(fpath)?
                 .par_lines()
                 .filter_map(|line| {
-                    matcher(&line).map(|(score, indices)| (line.into(), score, indices))
+                    matcher(&line).map(|(score, indices)| (line.to_string().into(), score, indices))
```
liuchengxu, line 65: `&str` -> `String`.

crates/stdio_server/src/session/handler/on_typed.rs
```diff
@@ -16,7 +17,7 @@ pub fn handle_on_typed(msg: Message, context: &SessionContext) {
 
     // TODO: sync for 100000, dyn for 100000+
     if let Some(ref source_list) = *source_list {
-        let source = filter::Source::List(source_list.iter().map(Into::into));
+        let source = filter::Source::List(source_list.iter().map(|s| s.to_string().into()));
```
liuchengxu, line 20: `&str` -> `String`.

PR #622, opened 2021-01-23
Add a bonus to the recently opened files

### LINKED ISSUE #542: Sort file list by hits, not alphabetical
CtrlP does that too... when you've visited a file before via the menu, it gets a score and will be shown first if the match is as good as the others. This is also the behavior of many other tools, such as rofi. It just means I spend less time looking for the file I actually want.

### LINKED ISSUE #563: Improve file searching: add bonus for recently opened files
**Problem**

I use `:Clap files` as my generic switch-to-file. Often I'll switch between recent files more than between recent and old files. I would be nice if the searching algorithm could give a bonus to recent files, instead of other less intuitive bonuses (eg shorter match).

**Solution**

Maintain a list of recent files and give a bonus to them.

**Other comments**

This behavior could be generalised for all other providers, that a recently selected item should be given a bonus if its score is the same as other matches.

This suggestion is also related to having a MRU provider, one that works and isn't based on the unreliable `:oldfiles`. I found a nice implementation here: https://github.com/markwu/vim-mrufiles

liuchengxu:
This PR adds a bonus (1/3 of the initial score) to the recently opened files tracked by `g:__clap_buffers`. https://github.com/liuchengxu/vim-clap/blob/498d8b34751300b1f5e763937b9525ef260fe670/plugin/clap.vim#L48

But it's neither persistent nor well-ordered as I hesitate to write them to the disk and do some more calculations on these BufEnter hooks.

Close #542 
Close #563

cc @romgrk @hasufell 

Before opening `autoload/clap/filter.vim`:

Screenshot from 2021-01-23 11-58-44

After opening `autoload/clap/filter.vim`:

Screenshot from 2021-01-23 11-59-09


romgrk:
Sorry for the absence, I've been swamped with work & personal issues. I've been reading your last PRs though, but I'm still catching up with all my notifications.

Haven't read the whole thing but just one question, this stores the MRU list by buffer number, not by file location, right? I wonder if it would be better to use the filepath, this could allow external plugins to save/restore the list in an easier way (eg [vim-mrufiles](https://github.com/markwu/vim-mrufiles)).

liuchengxu:
@romgrk We convert the buffer number to the full filepath when we need the list.  https://github.com/liuchengxu/vim-clap/blob/d7bef6a994dd35eef16168a06a73c758f93bf43b/autoload/clap/util.vim#L141-L143

PR #639, opened 2021-02-16
Support heuristic language aware bonus

liuchengxu:
For example, when we search for something, especially when searching in a buffer(`blines` provider), the target is more likely part of a function or variable declaration, we can add a positive bonus to these lines that can have a declaration. The rules are as follows:

- Add a positive bonus if the line can have a declaration.
- Add a negative bonus if the line is a comment line.

Any ideas are welcome, the current bonus rules are just my random thoughts.


romgrk:
The issue is basically where to get that information from. A few ideas:
 - vim: use syntax HL groups? problem is the file needs to be open, and the data needs to be passed to the backend
 - neovim: use lua's integrated treesitter information? problem: same as above
 - treesitter: integrate it in the rust backend? daunting task but once it's done the language heuristics problem is solved for good
 - tagfiles: existing tagfiles contain information about important lines (functions, vars, etc) for all supported languages. you also have the option to pass the files through a ctags program in case tagfiles dont exist

liuchengxu:
@romgrk Hmm, I think you misunderstand my intention of this PR, your ideas are basically unrelated. 

The idea of this PR is to add another bonus to the results returned from the base match algorithm(`fzy` etc) by checking if the line is possibly a declaration/comment or something. The checking rule is deadly simple for this version, we merely need to see if the line starts with `function` if we are searching in the context of vim source file, I think it's a good addition to `blines` provider.

It only takes effect when you specify the language context explicitly, and is only enabled for `blines` provider for now.

romgrk:
That's what I had understood, for `blines` using information from ctags it would be possible to avoid implementing custom checkers for each language (eg match `function` for vim files) because ctags already has all those regexes for a ton of languages.

But maybe I'm unclear or it's not possible with your implementation (I've just skimmed over the code), anyway this is a good start and will certainly improve results :)

liuchengxu:
Ok, I see you now. Yeah, this is the beginning and we can improve it later, apart from ctags, https://github.com/jacktasia/dumb-jump/blob/master/dumb-jump.el can also be a reference, but I believe we have to do some tweaks in clap.

PR #675, opened 2021-05-01
Handle the non-utf8 line in ripgrep's output

### LINKED ISSUE #673: Missing results from rg
If a source code line contains a unicode chars in comment, rg will return something like this:
```
{"type":"begin","data":{"path":{"text":".\\APP\\app_code\\SC.h"}}}
{"type":"match","data":{"path":{"text":".\\APP\\app_code\\SC.h"},"lines":{                                             "bytes":"xxxxxx"},"line_number":24,"absolute_offset":486,"submatches":[{"match                                             ":{"text":"xxxxxx"},"start":9,"end":23}]}}
{"type":"match","data":{"path":{"text":".\\APP\\app_code\\SC.h"},"lines":{                                             "bytes":"xxxxx"},"line_number":45,"absolute_offset":1751,"submatches":[{"                                             match":{"text":"xxx"},"start":9,"end":23}]}}
```
and clap will ignore this item.

liuchengxu:
Close #673

Previously, the non-utf8 line is simply ignored and this is not right, we should display it in any case.

Borrowed `bytelines` ideas from [#423](https://github.com/liuchengxu/vim-clap/pull/423/files#diff-422bd4801e1d35fe32f9a3dab487510fcf67b390f3b0b5fb25b0b6a23640d6c7), feel free to take a review @ImmemorConsultrixContrarie  :(.


crates/utility/src/bytelines.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn new(text: &'a [u8]) -> Self {
+        Self { text }
+    }
+}
+
+/// Newline char.
+const NL: u8 = b'\n';
+
+impl<'a> Iterator for ByteLines<'a> {
+    type Item = Line<'a>;
```
ImmemorConsultrixContrarie, line 43: I think it is easier to use `Cow<'a, str>` here.

crates/utility/src/bytelines.rs
```diff
@@ -0,0 +1,109 @@
...
+                // This line is the last one
+                self.text = &[];
+                text
+            }
+        };
+
+        Some(std::str::from_utf8(line).map_or(
+            Line::NotUtf8(String::from_utf8_lossy(line).to_string()),
+            Line::Utf8,
+        ))
```
ImmemorConsultrixContrarie, line 76: `map_or` eagerly evaluates its default value, thus here you always do `String::from_utf8_lossy(line).to_string()`, no matter what. Also, `to_string` is totally useless here and allocates one more string than you need. If you'd use `Cow` this should look like

```rust
Some(match std::str::from_utf8(line) {
    Ok(s) => s.into(),
    Err(_) => String::from_utf8_lossy(line).into(),
})
```

crates/utility/src/bytelines.rs
```diff
@@ -0,0 +1,109 @@
...
```
ImmemorConsultrixContrarie, line 105: Same thing about eager evaluation of `map_or` here.

crates/maple_cli/src/tools/ripgrep/jsont.rs
```diff
@@ -0,0 +1,100 @@
...
+    },
+}
+
+impl Data {
+    pub fn text(&self) -> String {
+        match self {
+            Self::Text { text } => text.to_string(),
+            Self::Bytes { bytes } => String::from_utf8_lossy(bytes).to_string(),
+        }
+    }
```
ImmemorConsultrixContrarie, line 80: `String::from_utf8_lossy(bytes).to_string()` allocates one more string than you need. Also, `Cow` here should be more effective, especially if most of `Data` would be text:

```rust
pub fn text(&self) -> Cow<str> {
    match self {
        Self::Text { text } => text.as_str().into(),
        Self::Bytes { bytes } => String::from_utf8_lossy(bytes).into(),
    }
}
```

Then, if you'd need a `String` from `Cow`, it has a `Cow::into_owned` function.

ImmemorConsultrixContrarie (commented):
I dunno about review, there's just too much unfamiliar code for me. Anyway, here's mini optimizations.

crates/maple_cli/src/tools/ripgrep/util.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+impl NiceDuration {
+    /// Returns the number of seconds in this duration in fraction form.
+    /// The number to the left of the decimal point is the number of seconds,
+    /// and the number to the right is the number of milliseconds.
+    fn fractional_seconds(&self) -> f64 {
+        let fractional = (self.0.subsec_nanos() as f64) / 1_000_000_000.0;
+        self.0.as_secs() as f64 + fractional
```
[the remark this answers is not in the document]
liuchengxu: I simply copy and paste this from the codebase of ripgrep https://github.com/BurntSushi/ripgrep/blob/3f4c4188c1/crates/printer/src/util.rs#L316, I guess it's fine to us.

crates/utility/src/bytelines.rs
ImmemorConsultrixContrarie: Oh, wait, I completely forgot that `String::from_utf8_lossy` returns `Cow`. The code here should be just
```rust
Some(String::from_utf8_lossy(line))
```

crates/utility/src/bytelines.rs
ImmemorConsultrixContrarie: Same thing here:
``rust
Some(String::from_utf8_lossy(line))
```

PR #689, opened 2021-05-27
accept complex query like command line program

QiBaobin:
it's not easy to send query like contain space, '+' etc. Let's use
a command line standard args syntax, all text after '--' is treated
as query instead of opts.

### DIFF
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -311,6 +311,7 @@ function! s:parse_opts(args) abort
   for arg in a:args
     if arg ==# '--'
       let g:clap.context.query = join(a:args[idx+1 :], ' ')
+      let idx = len(a:args)
       break
     endif
     if arg =~? '^++\w*=\w*'
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -312,7 +312,6 @@ function! s:parse_opts(args) abort
   for arg in a:args
     if arg ==# '--'
       let g:clap.context.query = join(a:args[idx+1 :], ' ')
-      let idx = len(a:args)
       break
     endif
     if arg =~? '^++\w*=\w*'
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -308,6 +308,7 @@ endif
 
 function! s:parse_opts(args) abort
   let idx = 0
+  let g:clap.provider.args = []
   for arg in a:args
     if arg ==# '--'
       let g:clap.context.query = join(a:args[idx+1 :], ' ')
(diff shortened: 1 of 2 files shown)


autoload/clap.vim
```diff
@@ -309,6 +309,11 @@ endif
 function! s:parse_opts(args) abort
   let idx = 0
   for arg in a:args
+    if arg ==# '--'
+      let g:clap.context.query = join(a:args[idx+1 :], ' ')
+      let idx = len(a:args)
+      break
+    endif
```
liuchengxu, line 316: Nice to have the docs updated since this patch changes how `query` is parsed and the way of users use it.

doc/clap-support.txt
```diff
@@ -97,6 +97,11 @@ Supported Providers & Tools                                    *clap-support*
                          the git base directory or `getcwd()` . For example,
                          use `Clap grep ..` to grep from the parent directory.
 
+                         Use `Clap grep -- word1 word2..` to grep a complex
+			 query, all text after -- would be the query text. You
+			 have to pass `++opt=[OPTION]` and `[DIR]` befor `--`
+			 if there is any of them.
```
liuchengxu, line 103:
```suggestion
                         Use `Clap grep -- word1 word2..` to grep a complex
			                   query, all text after -- would be treated as the query
			                   text. You have to pass `++opt=[OPTION]` and `[DIR]`
			                   before `--` if there is any of them.
```

autoload/clap.vim
```diff
@@ -308,7 +308,13 @@ endif
 
 function! s:parse_opts(args) abort
   let idx = 0
+  let g:clap.provider.args = []
   for arg in a:args
+    if arg ==# '--'
+      let g:clap.context.query = join(a:args[idx+1 :], ' ')
+      let idx = len(a:args)
```
liuchengxu, line 315:
```suggestion
```

QiBaobin:
Hey @liuchengxu , if we remove the line, all query would added to the provider.args too. do we want to do that? Currently, only text before `--` would be added.

```
      let idx = len(a:args)
```

liuchengxu:
>if we remove the line, all query would added to the provider.args too.

Why is that? You break the for loop when meeting `--`, making `let idx = len(a:args)` unused. Removing that line doesn't make any difference at all, the query is still the combination of items after `--`.

QiBaobin (commented):
> > if we remove the line, all query would added to the provider.args too.
> 
> Why is that? You break the for loop when meeting `--`, making `let idx = len(a:args)` unused. Removing that line doesn't make any difference at all, the query is still the combination of items after `--`.

You are correct, I forgot the change I made, thanks!

liuchengxu:
Merged, thank you!

PR #693, opened 2021-05-28
Add border and zindex for neovim's floating_win

liuchengxu:
This change requires a newer neovim since https://github.com/neovim/neovim/pull/14468 is just merged recently.

截屏2021-05-28 下午6 47 17


romgrk:
Is it possible to make the border optional? With the background shadow & different background colors it's not really necessary to have it.

liuchengxu:
@romgrk Thanks for the reminder, I have applied `g:clap_popup_border` for neovim's floating win as well, even though the name is not exactly precise.

romgrk:
Sounds good :)

PR #703, opened 2021-06-06
Added onehalf themes

kerunaru:
Based on the simple and clean [onehalf themes](https://github.com/sonph/onehalf), I created these for vim-clap. I put some comments specifying which color identifier I used for every Clap element.


liuchengxu:
Merged, thank you!

PR #729, opened 2021-08-01
Refactor tags command

liuchengxu:
The future plan is to support some features of https://github.com/universal-ctags/citre in vim-clap.

### DIFF
--- a/autoload/clap/maple/command.vim
+++ b/autoload/clap/maple/command.vim
@@ -124,7 +124,7 @@ function! clap#maple#command#tags(is_forerunner) abort
   let global_opts = has_key(g:clap.context, 'no-cache') ? ['--no-cache'] : []
   let global_opts = s:inject_icon_painter_opt(global_opts)
 
-  let subcommand = ['ctags', '', clap#rooter#working_dir()]
+  let subcommand = ['ctags', clap#rooter#working_dir()]
   if a:is_forerunner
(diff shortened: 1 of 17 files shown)


phongnh:
@liuchengxu because you're working on tags command, could you consider supporting custom/user ctags binary in `proj_tags` provider via settings or provider options? `Clap proj_tags` is using fixed `ctags`:

```vim
let s:support_json_format =
      \ len(filter(systemlist('ctags --list-features'), 'v:val =~# ''^json''')) > 0
```

Thanks.

liuchengxu:
@phongnh Not important for me as it's not hard to inject the binary into the PATH if you already have one, but I'm happy to accept PR.

phongnh:
@liuchengxu You're right. In my machine (Ubuntu), Universal Ctags is installed as `ctags-universal` instead of `ctags`, I'm not sure why (maybe conflicting with Exuberant Ctags). Maybe I just need to build and install Universal Ctags from source instead of package manager. Thanks.

romgrk:
@phongnh Or just create a symlink? Something like ``ln -s `which ctags-universal` /usr/local/bin/ctags``

phongnh:
> @phongnh Or just create a symlink? Something like `` ln -s `which ctags-universal` /usr/local/bin/ctags ``

Thank you @romgrk

PR #733, opened 2021-08-08
Resolve invoking Makefiles on FreeBSD/OpenBSD

### LINKED ISSUE #732: "Clap install" assumes a specific location for bash
While `bash` generally is found in `/usr/bin/bash`, on FreeBSD it is installed in `/usr/local/bin/bash`
This causes the installation of binaries to fail on FreeBSD.

Additionally, trying to make the python modules causes similar problem on FreeBSD as there are two `make` commands available: `make` and `gmake` (GNU)

If you point me in right direction, I may be able to put together a PR for these.

spamwax:
- `installer.vim` commands spawn a shell on *nix systems to build maple/python
  modules. We first ensure vim's `&shell` is set correctly and executable.
  Additionally, if we are on BSD system, we call `gmake` instead of make
  and pass an enviornment variable to Makefile.
  BSD users are informed if they do not have `gmake` installed on their
  system.
Resolves #732

### DIFF
--- a/autoload/clap/installer.vim
+++ b/autoload/clap/installer.vim
@@ -26,6 +26,7 @@ function! s:run_term(cmd, cwd, success_info) abort
     call termopen(a:cmd, {
           \ 'cwd': a:cwd,
           \ 'on_exit': {job, status -> s:OnExit(status)},
+          \ 'env': {'MAKE_CMD': s:make_cmd},
           \})
   else
     let cmd = a:cmd
@@ -44,6 +45,7 @@ function! s:run_term(cmd, cwd, success_info) abort
   noautocmd wincmd p
 endfunction
 
+let s:make_cmd = 'make'
 if has('win32')
   let s:from = '.\fuzzymatch-rs\target\release\fuzzymatch_rs.dll'
   let s:to = 'fuzzymatch_rs.pyd'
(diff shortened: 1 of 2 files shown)


autoload/clap/installer.vim
```diff
@@ -146,5 +151,23 @@ function! clap#installer#install(try_download) abort
...
 
+function! s:unix_sanity_check() abort
+  " If &shell is not set properly, everything will fail
+  if executable(&shell) != 1
+    call clap#helper#echo_error('Shell not executable. Check if '. &shell . 'exists!')
+  endif
+  " If on *BSD, we need to invoke gmake for clap's Makefiles
+  if ! has('win32') && executable('uname')
+    let l:uname = substitute(system('uname'), '\n', '', '')
+    if l:uname == 'FreeBSD' || l:uname == 'OpenBSD'
```
liuchengxu, line 162:
```suggestion
    if l:uname ==? 'FreeBSD' || l:uname ==? 'OpenBSD'
```

autoload/clap/installer.vim
```diff
@@ -146,5 +151,23 @@ function! clap#installer#install(try_download) abort
   endif
 endfunction
 
+function! s:unix_sanity_check() abort
+  " If &shell is not set properly, everything will fail
+  if executable(&shell) != 1
+    call clap#helper#echo_error('Shell not executable. Check if '. &shell . 'exists!')
+  endif
+  " If on *BSD, we need to invoke gmake for clap's Makefiles
+  if ! has('win32') && executable('uname')
```
liuchengxu, line 160:
```suggestion
  if !has('win32') && executable('uname')
```

autoload/clap/installer.vim
```diff
@@ -102,6 +106,7 @@ function! clap#installer#download_binary() abort
   if has('win32')
     let cmd = 'Powershell.exe -ExecutionPolicy ByPass -File "'.s:plugin_root_dir.'\install.ps1"'
   else
+    call s:unix_sanity_check()
```
liuchengxu, line 109: It could better to just stop the following execution if the check fails:

```vim
if !s:unix_sanity_check_is_ok()
  return
endif
```

spamwax: Thought that `clap#helper#echo_error` aborts the call :) Just pushed a new commit.

liuchengxu:
Merged, thank you!

PR #738, opened 2021-08-14
Support fzf search syntax

### LINKED ISSUE #492: Support fzf like search syntax
See what the fzf like search syntax is in https://github.com/junegunn/fzf#search-syntax.

This is a stand-alone task that given a raw search request and a line, that matcher should return the matched items accordingly. A rework for `vim-clap/crates/matcher` is required.

References:

- https://github.com/junegunn/fzf
- The Rust version of origin fzf: https://github.com/lotabout/skim

liuchengxu:
This PR implements most of [the fzf-like search syntax](https://github.com/junegunn/fzf#search-syntax). Happy to have some Rust suggestions if you happen to get some time :P @ImmemorConsultrixContrarie .

TODO:

- ~[ ] OR operator.~ I decide this won't be implemented as you can use multiple exact matches instead.

Close #492

截屏2021-08-14 上午11 24 57


fannheyward:
@liuchengxu I've rebuild the maple bin, but still can't search with the syntax:

截屏2021-08-17 上午9 28 58

fannheyward:
Does the Python module is must-have?

zhuzhzh:
can we support fzf-like search by default?  now it needs `let g:clap_builtin_fuzzy_filter_threshold = 0`.

liuchengxu:
`let g:clap_builtin_fuzzy_filter_threshold = 0` now is required to ensure the search is using the Rust backend instead of the other implementations because the fzf search syntax support is only available when using the Rust binary. Perhaps we can consider making it a default behavior if the Rust executable exists.

Issue #846, opened 2022-06-13
Pre-built install error on maple self-update from v0.34 to v0.35


rene-descartes2021:
**Environment (please complete the following information):**
 - OS: Debian Bullseye
 - (Neo)Vim version: Vim 8.2
 - vim-clap version: v0.34->v0.35 plug update
 - Have you reproduced with a minimal vimrc: no, space-vim
 - Have you updated to the latest plugin version: yes, well trying to
 - Have you upgraded to/compiled the latest Rust binary: no this fails 

**Describe the bug**
vim-plug `:PlugUpdate` command results in:
```
x Post-update hook for vim-clap ... Vim(call):E631: channel_write_input(): write failed
```
*EDIT:* I think an edited out remark I had here earlier was from vim-lsp issue, not related.

From `:Clap install-binary!` command in:  `Plug 'liuchengxu/vim-clap', { 'do': ':Clap install-binary!' }`

Problem appears to be [this line](https://github.com/liuchengxu/vim-clap/blob/master/autoload/clap/installer.vim#L134). 

**To Reproduce**

1. Create the minimal vimrc `min.vim`:

```vim
set nocompatible
set runtimepath^=/path/to/vim-clap
syntax on
filetype plugin indent on
```

2. Start (neo)vim with command: `vim -u min.vim`

3. Reproducible with:
```
:echo term_start($HOME.'/.vim/plugged/vim-clap/bin/maple upgrade --do
wnload', {'curwin': 1, 'cwd': $HOME.'/.vim/plugged/vim-clap', 'exit_c
b': {job, status -> execute('echoe "v:shell_error="'.status)}})
:echo v:shell_error
```
Or:
```
:Clap install-binary!
```

Output of above is:
```
Retrieving the latest remote release info...
New maple release v0.35 is avaliable, downloading...
thread 'main' panicked at 'not currently running on a Tokio 0.2.x runtime.',
 /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.25/src/runt
ime/handle.rs:118:28
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrac
e
101
```

v:shell_error is 101

With `let $RUST_BACKTRACE=1`:
```
Retrieving the latest remote release info...
New maple release v0.35 is avaliable, downloading...
thread 'main' panicked at 'not currently running on a Tokio 0.2.x runtime.',
[...]

PR #847, opened 2022-06-13
Workaround v0.34->v0.35 maple self-update failure

rene-descartes2021:
A proposal to workaround bug #846.

When using the pre-built binary:
- When maple fails to self-update, attempts to download the updated release directly instead.

Notes:
- Made a callback to call from s:OnExit which is called when a:status indicates a failure. In the case of maple self-update failure it then tries to download directly.
- Had to move s:OnExit closure to outer script scope, because the nested call to s:run_term couldn't redefine s:OnExit closure while in use.
- This is a workaround for the problem, the bug in the maple self-update could be fixed too or instead.

### DIFF
--- a/autoload/clap/installer.vim
+++ b/autoload/clap/installer.vim
@@ -6,26 +6,28 @@ set cpoptions&vim
 
 let s:plugin_root_dir = fnamemodify(g:clap#autoload_dir, ':h')
 
-function! s:run_term(cmd, cwd, success_info) abort
+function! s:OnExit(status, bufnr, success_info, ErrorCallback) abort
+  if a:status == 0
+    execute 'silent! bd! '.a:bufnr
+    call clap#helper#echo_info(a:success_info)
+  else
+    call a:ErrorCallback()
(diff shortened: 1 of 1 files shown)


autoload/clap/installer.vim
```diff
@@ -6,26 +6,28 @@ set cpoptions&vim
 
 let s:plugin_root_dir = fnamemodify(g:clap#autoload_dir, ':h')
 
-function! s:run_term(cmd, cwd, success_info) abort
+function! s:OnExit(status, bufnr, success_info, Error_Callback) abort
```
liuchengxu, line 9: And the ones in other places.

```suggestion
function! s:OnExit(status, bufnr, success_info, ErrorCallback) abort
```

liuchengxu:
A minor nit, otherwise LGTM. Nice work!

autoload/clap/installer.vim
rene-descartes2021: Done

Issue #573, opened 2020-10-28
Scroll Preview


sphynx79:
Please can you implement or tell me how can i scroll the preview.

Thanks

polyzen:
https://github.com/liuchengxu/vim-clap#keybindings

sphynx79:
good but i don't want scroll the file list, but the preview file window

liuchengxu:
That can be done, but need a bit more work as the current preview implementation does not read the whole file but the neccessary lines for being performant. When you scroll the preview window, we have to keep reading more of the origin file.

comfortablynick:
This would be great, since fzf is able to do it (seemingly without a performance penalty).

liuchengxu:
The preview of vim-clap will also be enhanced once #798 is merged, which is to extract and show the current context symbol using ctags. Maybe helpful for those who are interested in this issue.

liuchengxu:
I finally find some time to implement this feature that I'm also looking forward to for a long time. Feel free to try out #933, don't forget to recompile the binary before the next release is available.

Issue #532, opened 2020-09-30
Highlighting with tree-sitter


BonaBeavis:
[Treesitter](https://github.com/nvim-treesitter/nvim-treesitter) does not seem to work with the previews. But would be a good fit, as you are very concerned with speed.

liuchengxu:
#1019 finally implements this feature after 3 years :P

Issue #235, opened 2020-01-13
Show history when input is empty


arashm:
**Describe the solution you'd like**
When doing `:Clap files` or `:Clap grep`, many times I find myself wanna do the same search as the one I already did to select a different result, for that I need to type the whole input again. It could be cool if Clap could show a list of 10 previous searches when the input is empty, and we could just select it from the list.

liuchengxu:
My original plan is to use `<C-N>`/`<C-P>` for navigating the input history, that could make type the same search easier.

svermeulen:
@liuchengxu is input history available yet as a feature?  `<c-n>` / `<c-p>` does not currently seem to do this

liuchengxu:
Still unimplemented, not a priority for me. PR welcome.

mtuzinskiy:
I miss this feature in vim-clap, would be great to have it

liuchengxu:
Sorry for taking so long on this issue, for the simplicity, a new provider can be added for such a purpose. Could you try this branch https://github.com/liuchengxu/vim-clap/tree/input-history-provider? @arashm @svermeulen @mtuzinskiy It's not complete, but we're starting somewhere anyway. Now you can invoke `:Clap input_history` to show the last 20 queries, and rerun the last _normal_(anyone but not `input_history` and `providers` provider) provider with that query. The future work could be make a decision if we want to show this history whenever the input is empty for any providers or just the `input_history` provider.

arashm:
That's great news @liuchengxu. Will try out that branch today 💪

arashm:
I recompiled the rust stuff, but I still get an empty window no matter what. Other providers I use, seems to be fine.

<img width="742" alt="Screenshot 2022-01-26 at 14 16 41" src="https://user-images.githubusercontent.com/849640/151179546-0d5a7c04-267c-4a4b-91ac-a2040be71e12.png">

liuchengxu:
The input history provider does not require the Rust binary. The idea is to record each input in this provider and use these history inputs as the source and it's not yet persistent, you need to call other providers with queries first and then call the input history provider to resume some query against the last provider.

The whole design is not settled. I'm happy to have some feedback on the UX.

liuchengxu:
@arashm I have pushed a new commit to fix the vim issue you just ran into, please pull the latest commit of that branch and try again. I didn't test Vim, only tested against NeoVim :(.

arashm:
[...]

PR #872, opened 2022-09-11
Rework the core logic of provider in Rust

### LINKED ISSUE #875: Use the library ripgrep instead of relying on the executable `rg`
ripgrep provides a library https://github.com/BurntSushi/ripgrep/blob/master/crates/grep/README.md which can be easily integrated into our Rust backend. The story is that while working on #872, I found that collecting the source items by invoking the shell command `rg` is somewhat annoying as the shell command does not produce items linearly, the output is empty from some point and then spikes suddenly, leading to a bad user experience. Responsiveness is really important. Maybe I did something wrong to collect the source item from the shell command, but with a quick experiment, I found using the crate `grep` helps resolve it and it'd also remove one explicit external binary dependency, hence I think it's a good addition.

Other potential benefits:

- Once the new search syntax is supported, e.g., use `+git` to include only the results from git tracked files, using the library ripgrep may help skip the unrelated files at the stage of walking the directory.

liuchengxu:
The goal of this PR is to overhaul the bridge between Vim and Rust, making it possible to delegate all the heavy jobs onto the Rust backend and make Vim a mere UI layer as much as possible. It's mainly achieved by allowing the Rust backend to retrieve and manipulate the Vim state on demand, which also unlocks a lot of features I'm looking forward to, such as tree-sitter highlights(#532), the preview scroll(#573), and the input history(#235).

Notable changes:

- Switch to the new stdio_server implementation.
    - Allow calling Vim methods actively, this is the key to support a lot of new features later.
    - A lot of old workarounds for improving the performance are able to be safely removed, and a bunch of vimscripts have been deleted.
    - **Now the Rust binary is mandatory to make each provider work** even if the source of it is purely from the inside Vim, in which case the source items will be passed to the Rust backend.
- More flexible debounce delay for on_typed hook.
- Add on_move debounce delay on the Rust side.
- Specify the `icon` attribute in each provider explicitly if it's supported, `Null` by default.
- Allow setting the `debounce` per provider, the debounce is enabled by default.
- The required Rust version is updated to the latest stable.

After switching to the native searcher instead of processing the output of the shell command `rg`, the `grep` and `files`, `blines` provider is significantly faster and more responsive.

There are still some TODOs and optimizations that can be done, but it should work in most cases, any feedback is welcome and testers are appreciated.

Close #875


rene-descartes2021:
Seems to work fine in my testing, with:
```bash
cd ~/.vim/plugged/vim-clap
git branch remove-deprecated-runner origin/remove-deprecated-runner
git co remove-deprecated-runner
vim
```
And in Vim:
```vim
call clap#installer#install(v:false)
```
`<space>pf` and `<space>pr` in space-vim work just fine for me in this vim-clap branch.

liuchengxu:
@rene-descartes2021 Thanks for the feedback! I'll probably hold this back until #875 is resolved as I found that the performance at present is okay to process hundreds of thousands of items, but not satisfying when using grep against a data set with millions of items, even if that's a rare case.

PR #878, opened 2022-09-30
make: adding 'install'

Shane-XB-Qian:


Shane-XB-Qian:
usage:
after `make`, then `make install` to copy the built to `bin/`, nothing else but just found I need to manually move that repeatly if no such.
// or perhaps there was better way, I just noticed this after 10 mins which did not look into very much. :-)

--
shane.xb.qian

Shane-XB-Qian:
ok, I see.
but seems a little tricky. and the built version would be not updated to be same if so or perhaps.
I supposed that better be to have a `make install`, and this was a usual way.

--
shane.xb.qian

liuchengxu:
The locally built binary is always preferred over the pre-built binary as the pre-built binary might be incompatible if some breaking changes are introduced even it's not frequent. `make install` sounds like `bin/` is the official place to put the binary, but it's actually not, we recommend users (re-)compiling the binary locally so that the latest features/fixes can be used once you pull the latest commit of the repo.

Shane-XB-Qian:
um, `make install` is a manual action, unless user manually kicked off such then would override the one in `bin/`.
// you can check fzf, his makefile had a 'install' and put it in 'bin/fzf' as well. :-)

liuchengxu:
The majority of fzf users don't compile the binary themselves IMHO, but we do recommend users compile the binary on their own, and fzf does not search the binary in other places to my knowledge, hence it's not exactly the same. In our case, simple `make` is enough, rendering this operation unnecessary. `make install` is a manual action, but this operation name is so typical that people may have a misunderstanding that the binary has to be installed somewhere.

All in all, I'm not sure it's a good addition, and it's not mentioned in neither README.md nor INSTALL.md.

Makefile
```diff
@@ -8,4 +8,7 @@ maple:
 python-dynamic-module:
 	cd pythonx/clap && $(MAKE_CMD) build
 
-.PHONY: all maple python-dynamic-module
+install:
```
liuchengxu, line 11: I'm reluctant to use `install`, can we use `bin/maple` or something else?

Shane-XB-Qian: never mind, it was just a proposal.

PR #929, opened 2023-02-13
Save and restore winminheight in addition to winheight

xorander00:
Save and restore winminheight in addition to winheight when opening and closing the floating window


liuchengxu:
Thank you!

PR #1079, opened 2024-05-09
Fix build for Android

truboxl:
Fix https://github.com/termux/termux-packages/issues/20063


truboxl:
I also dont know if it works on Android. Gotta give credit to @luisdavim

liuchengxu:
Let me know if I should publish a new release for this patch. @luisdavim BTW, I seem to observe some regression of UI hanging with the latest version locally, still looking into it. Please record the current version you are working with, and file an issue if you run into any problems.

luisdavim:
Awesome, thanks, I will give this a try and let you know.

PR #1105, opened 2024-12-12
Update variable name for source to avoid function name conflict

phongnh:
Hi @liuchengxu,

I am adding PR to update variable name from `Source` to `ClapProviderSource` to avoid function name conflict. In my case, I have a `Source` function to source vim config. I can change my config to fix this issue but I think that `Source` is quite common function name, others can be affected by this error as well. Please help me to review and accept this PR if you find it does make senses, thanks.

```
Error detected while processing function clap#[45]..clap#open_provider[23]..<SNR>45_detect_source_type:
line    1:
E705: Variable name conflicts with existing function: Source
```

### DIFF
--- a/autoload/clap.vim
+++ b/autoload/clap.vim
@@ -125,15 +125,15 @@ function! s:inject_default_impl_is_ok(provider_info) abort
 endfunction
 
 function! s:detect_source_type() abort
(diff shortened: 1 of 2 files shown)


liuchengxu:
Look good to me. CI failure is unrelated.

PR #1108, opened 2024-12-20
Use etcetera instead of directories

### LINKED ISSUE #1107: Could be possible to switch to use etcetera and use Xdg paths on macOS as well
[etcetera](https://docs.rs/etcetera/latest/etcetera/) provides a mean to get the Xdg paths on macOS as well giving a much easier life to people that share their home .config between linux and mac.

I can get you a patch if you thing it is acceptable.

lu-zero:
It provides more flexibility and it has a better default for cli tools.

Fixes #1107.

### DIFF
--- a/crates/maple_config/src/lib.rs
+++ b/crates/maple_config/src/lib.rs
@@ -14,10 +14,9 @@ fn load_config(
     specified_config_file: Option<PathBuf>,
 ) -> (Config, PathBuf, Option<toml::de::Error>) {
     let config_file = specified_config_file.unwrap_or_else(|| {
-        // Linux: ~/.config/vimclap/config.toml
-        // macOS: ~/Library/Application\ Support/org.vim.Vim-Clap/config.toml
+        // Linux & macOS: ~/.config/vimclap/config.toml
         // Windows: ~\AppData\Roaming\Vim\Vim Clap\config\config.toml
-        let config_file_path = Dirs::project().config_dir().join("config.toml");
+        let config_file_path = Dirs::config_dir().join("config.toml");
 
         if !config_file_path.exists() {
             std::fs::create_dir_all(&config_file_path).ok();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -748,19 +748,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "directories"
-version = "4.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f51c5d4ddabd36886dd3e1438cb358cdcb0d7c499cb99cb4ac2e38e18b5cb210"
-dependencies = [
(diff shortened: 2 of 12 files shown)


crates/maple_config/src/lib.rs
```diff
@@ -14,10 +14,9 @@ fn load_config(
     specified_config_file: Option<PathBuf>,
 ) -> (Config, PathBuf, Option<toml::de::Error>) {
     let config_file = specified_config_file.unwrap_or_else(|| {
-        // Linux: ~/.config/vimclap/config.toml
-        // macOS: ~/Library/Application\ Support/org.vim.Vim-Clap/config.toml
+        // Linuxi & macOS: ~/.config/vimclap/config.toml
         // Windows: ~\AppData\Roaming\Vim\Vim Clap\config\config.toml
-        let config_file_path = Dirs::project().config_dir().join("config.toml");
+        let config_file_path = Dirs::config_dir().join("config.toml");
```
liuchengxu, line 19: Can we keep the support of ` ~/Library/Application\ Support/org.vim.Vim-Clap/config.toml` on macOS? This will ensure compatibility for existing macOS users, preventing any complaints about their current configuration file no longer working.

crates/maple_config/src/lib.rs
```diff
@@ -14,10 +14,9 @@ fn load_config(
     specified_config_file: Option<PathBuf>,
 ) -> (Config, PathBuf, Option<toml::de::Error>) {
     let config_file = specified_config_file.unwrap_or_else(|| {
-        // Linux: ~/.config/vimclap/config.toml
-        // macOS: ~/Library/Application\ Support/org.vim.Vim-Clap/config.toml
+        // Linuxi & macOS: ~/.config/vimclap/config.toml
```
liuchengxu, line 17:
```suggestion
        // Linux & macOS: ~/.config/vimclap/config.toml
```

crates/maple_config/src/lib.rs
lu-zero: I can add the logic to look on both places.

crates/maple_config/src/lib.rs
lu-zero: Oh sure ^^;

lu-zero:
This will have to wait until https://github.com/lunacookies/etcetera/pull/22 lands and is released if you want to stay compatible.

liuchengxu:
The upstream PR has been pending for quite a while, I don't mind using a patched dep of etcetera since it seems to be stable if you would like this PR merged into vim-clap sooner. CI failure is unrelated and can be ignored.

lu-zero:
I wished to not bring in another custom version of the same logic, if only the `directories` author weren't that stubborn...

utkarshgupta137:
v0.9 released with fix

